PR #1, opened 2015-02-18
COL-23 - Set up core Collabosphere infrastructure

nicolaasmatthijs:
- Basic module structure
- Express app server
- Logging
- Configuration
- Travis build
- Test framework

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,58 @@
+/**
+ * Copyright 2015 UC Berkeley (UCB) Licensed under the
+ * Educational Community License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may
+ * obtain a copy of the License at
+ *
+ *     http://opensource.org/licenses/ECL-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS"
+ * BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+var gulp = require('gulp');
+var jscs = require('gulp-jscs');
+var mocha = require('gulp-mocha');
+var runSequence = require('run-sequence');
+
+/**
+ * Run the JSCS code style linter
+ */
+gulp.task('jscs', function() {
+  return gulp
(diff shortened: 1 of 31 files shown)


nicolaasmatthijs:
I'll be cleaning this up a bit more, but this is ready for an initial review pass

gulpfile.js
```diff
@@ -0,0 +1,58 @@
...
+gulp.task('mocha', function() {
+  return gulp
+    .src(['node_modules/col-tests/lib/beforeTests.js', 'node_modules/col-*/tests/**/*.js'])
+    .pipe(mocha());
+});
+
+/**
+ * Run the full Collabosphere test suite
+ */
+gulp.task('test', function() {
```
simong: Can we make this the default gulp task?

node_modules/col-core/lib/server.js
```diff
@@ -0,0 +1,257 @@
...
+  app.set('json spaces', 0);
+
+  // Don't output the x-powered-by header
+  app.set('x-powered-by', false);
+
+  /*!
+   * Collabosphere supports the following type of request encodings:
+   *
+   *  * urlencoded (regular POST requests)
+   *  * application/json
```
simong: Is there a need for a JSON or multipart parser? It seems like most of the API interactions with this tool should be doable through plain-old urlencoded requests right?

simong:
LGTM, are you planning on removing that dummy service and putting something more tangible before you want to get this merged?

gulpfile.js
nicolaasmatthijs: I was thinking about making the production build the default task

node_modules/col-core/lib/server.js
nicolaasmatthijs: Multipart will be necessary, as students will be able to upload files to the asset library. We can probably take out the application/json encoding though.

nicolaasmatthijs:
@simong : I was thinking I would just leave the dummy service to give people a better idea of how it will work. As soon as the first service is in, we can take it out.

nicolaasmatthijs:
Followed up on review. Back to @simong for review.

node_modules/col-core/lib/server.js
simong: I can't recall the designs immediately but would they upload into the tool directly or as part of an assignment?

nicolaasmatthijs: Both

PR #2, opened 2015-02-19
Added the initial model

simong:
This adds in the model as sketched out yesterday. It currently supports:
 - Canvas
 - Courses
 - Users
 - Activities (and their types)
 - Assets (and their categories)
 - Comments

### DIFF
--- a/node_modules/col-core/lib/db.js
+++ b/node_modules/col-core/lib/db.js
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
 var setUpModel = function(sequelize) {
 
   /**
-   * The Course model
+   * The `canvas` table keeps track of the Canvas instances that Collabosphere should
+   * communicate with. It holds all the information that's needed to interact with the
+   * Canvas API and to embed on of the Collabosphere tools into Canvas
    *
-   * @property  {Number}      id                      The id of the course
+   * @property  {String}    app_domain      The domain where canvas is running on
+   * @property  {String}    api_key         The key that can be used to interact with the Canvas API
+   * @property  {String}    lti_key         The basic LTI key that will be used to embed the tools into Canvas
+   * @property  {String}    lti_secret      The basic LTI secret that will be used to embed the tools into Canvas
+   * @property  {Boolean}   use_https       Whether the Canvas API is running on https
    */
-  var Course = module.exports.Course = sequelize.define('Course', {
-    'id': {
+  var Canvas = module.exports.Canvas = sequelize.define('canvas', {
+    'app_domain': {
+      'type': Sequelize.STRING,
+      'allowNull': false,
+      'primaryKey': true
+    },
+    'api_key': {
+      'type': Sequelize.STRING,
+      'allowNull': false,
+    },
+    'lti_key': {
+      'type': Sequelize.STRING,
+      'allowNull': false,
+      'unique': true
+    },
+    'lti_secret': {
+      'type': Sequelize.STRING,
+      'allowNull': false,
+      'unique': true
+    },
+    'use_https': {
+      'type': Sequelize.BOOLEAN,
+      'defaultValue': true
+    }
+  }, {
+    'underscored': true,
+  });
+
+  /**
+   * The `course` table keeps track of each course where the tools have been embedded in
+   *
+   * @property  {Number}      canvas_course_id        The id of the course in Canvas
+   * @property  {Boolean}     enable_upload           Whether or not students are allowed to upload assets directly
+   * @property  {Boolean}     enable_sync             Whether or not this course's data should be synced with Canvas
+   */
+  var Course = module.exports.Course = sequelize.define('course', {
+    'canvas_course_id': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false
+    },
+    'enable_upload': {
+      'type': Sequelize.BOOLEAN,
+      'allowNull': false,
+      'defaultValue': true
+    },
+    'enable_sync': {
+      'type': Sequelize.BOOLEAN,
+      'allowNull': false,
+      'defaultValue': true
+    }
+  }, {
+    'underscored': true,
+  });
+
+  // Each course belongs to a specific Canvas instance
+  Course.belongsTo(Canvas, {'onDelete': 'CASCADE'});
+
+  /**
+   * The `user` table keeps track of each user that is enrolled in a course. This means that
+   * the same physical student might end up with two student records within Collabosphere. As
+   * there's no overlap between courses, this is acceptable.
+   *
+   * @property  {Number}        canvas_user_id          The id of the user in Canvas
+   * @property  {String}        canvas_course_role      The role of the user in the course
+   * @property  {String}        canvas_full_name        The full name of the student
+   * @property  {String}        canvas_image            The URL that points to an image for the user
+   * @property  {Number}        points                  The total points that the user has accumulated in the course
+   * @property  {Boolean}       share_points            Whether the user wants to share their point total with the other users in the course
+   */
+  var User = module.exports.User = sequelize.define('user', {
+    'canvas_user_id': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false
+    },
+    'canvas_course_role': {
+      'type': Sequelize.STRING,
+      'allowNull': false
+    },
+    'canvas_full_name': {
+      'type': Sequelize.STRING,
+      'allowNull': false,
+    },
+    'canvas_image': {
+      'type': Sequelize.STRING,
+      'allowNull': true,
+    },
+    'points': {
       'type': Sequelize.INTEGER,
       'allowNull': false,
-      'unique': true,
-      'primaryKey': true
+      'defaultValue': 0
+    },
+    'share_points': {
+      'type': Sequelize.BOOLEAN,
+      'allowNull': true,
+      'defaultValue': null
+    }
+  }, {
+    'underscored': true,
+  });
+
+  // Each user belongs to a specific course
+  User.belongsTo(Course, {'onDelete': 'CASCADE'});
+
+  // TODO: Define these and abstract them out in constants
+  var activityTypes = ['like', 'dislike', 'comment', 'submit_assignment'];
+  var objectTypes = ['canvas_submission', 'canvas_discussion', 'asset', 'comment'];
+
+  /**
+   * The `activity` table will keep track of actions that users make in the system. This can
+   * include things such as:
+   *  - liking or disliking assets
+   *  - commenting on assets
+   *  - submitting an assignment
+   *  - posting to a Canvas discussion
+   *  - creating a board
+   *  - ...
+   *
+   * As an activity can be related to different things, it's hard to enforce referential integrity without
+   * adding lots of unused foreign keys to the table. That's why a more general approach has been taken:
+   *  - `object_id` will be an identifier (e.g., canvas assignment id)
+   *  - `object_type` will hold information about the type the id points to (e.g., a canvas assignment)
+   *
+   * @property  {String}        type                The type of the activity
+   * @property  {Number}        object_id           The id of the object that was involved in the activity. This can be a Canvas assignment id, a comment id, an asset id, etc
+   * @property  {String}        object_type         The type of object that was involved in the activity
+   */
+  var Activity = module.exports.Activity = sequelize.define('activity', {
+    'type': Sequelize.ENUM(activityTypes),
+    'object_id': Sequelize.INTEGER,
+    'object_type': Sequelize.ENUM(objectTypes)
+  }, {
+    'underscored': true,
+  });
+
+  // Each activity happens within a course
+  Activity.belongsTo(Course, {'onDelete': 'CASCADE'});
+
+  // The user who triggered the activity
+  Activity.belongsTo(User, {'as': 'actor', 'onDelete': 'CASCADE'});
+
+  // The target is the user who was indirectly involved in the activity. For example,
+  // when John likes a comment from Bella, John will be the actor and Bella will be the target.
+  // When users drop out of a course they will be removed from the system. That doesn't
+  // mean however that people who liked or commented on assets generated by those users
+  // should lose points in their engagement index. That's why target ids will be set to
+  // null when such a user is removed from the system
+  Activity.belongsTo(User, {'as': 'target', 'onDelete': 'SET NULL'});
+
+  /**
+   * The `activity_type` table will keep track of the different activities that can occur
+   * and the weights that are associated with them.
+   *
+   * @property  {String}        type                The type of activity. For example `like`, `comment`, ..
+   * @property  {String}        name                The human-readable name of an activity. For example `Like`, `Comment`, ..
+   * @property  {Number}        points              The number of points this activity adds towards a user's engagement index
+   * @property  {Boolean}       enabled             Whether activities of this type should be added towards a user's engagement index
+   */
+  var ActivityType = module.exports.ActivityType = sequelize.define('activity_type', {
+    'type': {
+      'type': Sequelize.ENUM(activityTypes),
+      'allowNull': false
+    },
+    'name': {
+      'type': Sequelize.STRING,
+      'allowNull': false
+    },
+    'points': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false
+    },
+    'enabled': {
+      'type': Sequelize.BOOLEAN,
+      'allowNull': false
+    }
+  }, {
+    'underscored': true,
+  });
+
+  // Each activity type belongs to a specific course
+  ActivityType.belongsTo(Course, {'onDelete': 'CASCADE'});
+
+  // Each activity has one activity type
+  Activity.belongsTo(ActivityType, {'onDelete': 'CASCADE'});
+
+  // TODO: Define these and abstract them out in constants
+  var assetTypes = ['file', 'url', 'board', 'thought'];
+
+  /**
+   * The `asset` table will keep track of each asset that needs to be displayed in the asset library
+   *
+   * @property {Number}     canvas_assignment_id      The id of the assignment if the asset was submitted as part of an assignment in canvas
+   * @property {String}     type                      The type of asset
+   * @property {String}     title                     The title of the asset
+   * @property {String}     description               The description of the asset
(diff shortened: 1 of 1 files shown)


christianv:
@simong looks like `jscs` is having some trouble.

`npm install -g jscs`
`jscs .`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
 var setUpModel = function(sequelize) {
 
   /**
-   * The Course model
+   * The `canvas` table keeps track of the Canvas instances that Collabosphere should
+   * communicate with. It holds all the information that's needed to interact with the
+   * Canvas API and to embed on of the Collabosphere tools into Canvas
```
nicolaasmatthijs: `to embed the`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
    *
-   * @property  {Number}      id                      The id of the course
+   * @property  {String}    app_domain      The domain where canvas is running on
```
nicolaasmatthijs: Should this be `canvas_api_domain`. I think Canvas is giving us `api_domain` instead of `app_domain`, and it would make it sound less like the "domain on which the Collabosphere app is running"

nicolaasmatthijs: `The domain on which`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+    'use_https': {
+      'type': Sequelize.BOOLEAN,
+      'defaultValue': true
+    }
+  }, {
+    'underscored': true
+  });
+
+  /**
+   * The `course` table keeps track of each course where the tools have been embedded in
```
nicolaasmatthijs: `each course in which one of the tools has been embedded`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   *
+   * @property  {Number}      canvas_course_id        The id of the course in Canvas
+   * @property  {Boolean}     enable_upload           Whether or not students are allowed to upload assets directly
+   * @property  {Boolean}     enable_sync             Whether or not this course's data should be synced with Canvas
```
nicolaasmatthijs: Maybe this should just be `active`?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+    }
+  }, {
+    'underscored': true
+  });
+
+  /**
+   * The `course` table keeps track of each course where the tools have been embedded in
+   *
+   * @property  {Number}      canvas_course_id        The id of the course in Canvas
+   * @property  {Boolean}     enable_upload           Whether or not students are allowed to upload assets directly
```
nicolaasmatthijs: Can remove the `or not` part as per the docs above

nicolaasmatthijs: `upload to the Asset Library directly`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+  // Each course belongs to a specific Canvas instance
+  Course.belongsTo(Canvas, {'onDelete': 'CASCADE'});
+
+  /**
+   * The `user` table keeps track of each user that is enrolled in a course. This means that
+   * the same physical student might end up with two student records within Collabosphere. As
+   * there's no overlap between courses, this is acceptable.
+   *
+   * @property  {Number}        canvas_user_id          The id of the user in Canvas
+   * @property  {String}        canvas_course_role      The role of the user in the course
```
nicolaasmatthijs: Note that Canvas returns an array of roles for a user.

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * @property  {String}        canvas_full_name        The full name of the student
```
nicolaasmatthijs: We probably need the sortable name in here as well ?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * @property  {String}        canvas_image            The URL that points to an image for the user
```
nicolaasmatthijs: `to a profile picture for the user`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * @property  {Number}        points                  The total points that the user has accumulated in the course
```
nicolaasmatthijs: `Engagement Index points`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * @property  {Boolean}       share_points            Whether the user wants to share their point total with the other users in the course
```
nicolaasmatthijs: `other students`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+    }
+  }, {
+    'underscored': true
+  });
+
+  // Each user belongs to a specific course
+  User.belongsTo(Course, {'onDelete': 'CASCADE'});
+
+  // TODO: Define these and abstract them out in constants
+  var activityTypes = ['like', 'dislike', 'comment', 'submit_assignment'];
```
nicolaasmatthijs: There are a few others like creating a discussion and posting to a discussion

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+  var objectTypes = ['canvas_submission', 'canvas_discussion', 'asset', 'comment'];
+
+  /**
+   * The `activity` table will keep track of actions that users make in the system. This can
+   * include things such as:
+   *  - liking or disliking assets
+   *  - commenting on assets
+   *  - submitting an assignment
+   *  - posting to a Canvas discussion
+   *  - creating a board
```
nicolaasmatthijs: `whiteboard`

node_modules/col-core/lib/db.js
simong: Hence the `TODO`. I'll come back to this once the module structure has been set up. We can look into abstracting these out into proper constants

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * adding lots of unused foreign keys to the table. That's why a more general approach has been taken:
+   *  - `object_id` will be an identifier (e.g., canvas assignment id)
+   *  - `object_type` will hold information about the type the id points to (e.g., a canvas assignment)
+   *
+   * @property  {String}        type                The type of the activity
+   * @property  {Number}        object_id           The id of the object that was involved in the activity. This can be a Canvas assignment id, a comment id, an asset id, etc
+   * @property  {String}        object_type         The type of object that was involved in the activity
+   */
+  var Activity = module.exports.Activity = sequelize.define('activity', {
+    'type': Sequelize.ENUM(activityTypes),
```
nicolaasmatthijs: Should we specify `allowNull` for all of these?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+  // The user who triggered the activity
+  Activity.belongsTo(User, {'as': 'actor', 'onDelete': 'CASCADE'});
+
+  // The target is the user who was indirectly involved in the activity. For example,
+  // when John likes a comment from Bella, John will be the actor and Bella will be the target.
+  // When users drop out of a course they will be removed from the system. That doesn't
+  // mean however that people who liked or commented on assets generated by those users
+  // should lose points in their engagement index. That's why target ids will be set to
+  // null when such a user is removed from the system
+  Activity.belongsTo(User, {'as': 'target', 'onDelete': 'SET NULL'});
```
nicolaasmatthijs: Do we need to specify that the target can be null ?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+  });
+
+  // Each activity happens within a course
+  Activity.belongsTo(Course, {'onDelete': 'CASCADE'});
+
+  // The user who triggered the activity
+  Activity.belongsTo(User, {'as': 'actor', 'onDelete': 'CASCADE'});
+
+  // The target is the user who was indirectly involved in the activity. For example,
+  // when John likes a comment from Bella, John will be the actor and Bella will be the target.
```
nicolaasmatthijs: This should be `Christina` ;)

node_modules/col-core/lib/db.js
simong: The `SET NULL` kinda implies that no?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+  // When users drop out of a course they will be removed from the system. That doesn't
```
nicolaasmatthijs: `When users drop out of a course, they will be removed from Collabosphere. However, the users that received points through likes and comments from these users should not lose points in the Engagement Index. Therefore, the target ids will be set to null when such a user is removed from Collabosphere.`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+  // should lose points in their engagement index. That's why target ids will be set to
+  // null when such a user is removed from the system
+  Activity.belongsTo(User, {'as': 'target', 'onDelete': 'SET NULL'});
+
+  /**
+   * The `activity_type` table will keep track of the different activities that can occur
+   * and the weights that are associated with them.
+   *
+   * @property  {String}        type                The type of activity. For example `like`, `comment`, ..
+   * @property  {String}        name                The human-readable name of an activity. For example `Like`, `Comment`, ..
```
nicolaasmatthijs: Should we make this `displayname` ?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * @property  {Number}        points              The number of points this activity adds towards a user's engagement index
```
nicolaasmatthijs: `contributes` here and the line below

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+    },
+    'name': {
+      'type': Sequelize.STRING,
+      'allowNull': false
+    },
+    'points': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false
+    },
+    'enabled': {
```
nicolaasmatthijs: We can probably default this to `true`

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * The `activity_type` table will keep track of the different activities that can occur
+   * and the weights that are associated with them.
+   *
+   * @property  {String}        type                The type of activity. For example `like`, `comment`, ..
+   * @property  {String}        name                The human-readable name of an activity. For example `Like`, `Comment`, ..
+   * @property  {Number}        points              The number of points this activity adds towards a user's engagement index
+   * @property  {Boolean}       enabled             Whether activities of this type should be added towards a user's engagement index
+   */
+  var ActivityType = module.exports.ActivityType = sequelize.define('activity_type', {
+    'type': {
```
nicolaasmatthijs: We could also hold a default points configuration object in the codebase and only store the overrides in this table. In that case, the display name could probably just be part of the default configuration and we wouldn't have to store it in the database.

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+
+  /**
+   * The `asset` table will keep track of each asset that needs to be displayed in the asset library
+   *
+   * @property {Number}     canvas_assignment_id      The id of the assignment if the asset was submitted as part of an assignment in canvas
+   * @property {String}     type                      The type of asset
+   * @property {String}     title                     The title of the asset
+   * @property {String}     description               The description of the asset
+   * @property {String}     thumbnail_url             The thumbnail url of the asset
+   * @property {String}     large_url                 The large url of the asset
```
nicolaasmatthijs: Should we add `[]` indicators for the optional fields?

node_modules/col-core/lib/db.js
simong: It returns a comma-separated string. I was thinking the `User` model would have an instance function that allows you to determine whether a user can manage a role.
Something to the tune of:

```javascript
var User = define('user', { ... }, {
  'instanceMethods': {
    /**
     * @returns {Boolean}       Whether a user is an instructor
     */
    'isInstructor': function() {
       return this.dataValues.canvas_course_role.indexOf('Instructor');
    },

    /**
     * @returns {Boolean}       Whether a user is a TA
     */
    'isTA': function() {
       return this.dataValues.canvas_course_role.indexOf('TA');
    },

    /**
     * @returns {Boolean}       Whether a user is an administrator
     */
    'isAdministrator': function() {
       return this.dataValues.canvas_course_role.indexOf('Administrator');
    }

    /**
     * @returns {Boolean}       Whether the user can manage the associated course
     */
    'canManageCourse': function() {
       return (this.isInstructor() || this.isTA() || this.isAdministrator());
    }
  }
});
```

That should make things like doing authz checks for the engagement index / points configuration fairly straightforward

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * @property {String}     url                       The url of the asset
+   * @property {String}     source                    The source of the asset
+   * @property {String}     body                      The body of the asset
+   * @property {Number}     likes                     The number of people who liked the asset
+   * @property {Number}     dislikes                  The number of people who disliked the asset
+   * @property {Number}     views                     The number of times the assets has been viewed
```
nicolaasmatthijs: typo

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+   * @property {Number}     comments                  The number of people who commented on the asset
```
nicolaasmatthijs: Maybe we can make this `comment_count`?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+      'type': Sequelize.INTEGER,
+      'allowNull': true
+    },
+    'type': {
+      'type': Sequelize.ENUM(assetTypes),
+      'allowNull': true
+    },
+    'title': {
+      'type': Sequelize.STRING,
+      'allowNull': true
```
nicolaasmatthijs: Title can't be null

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+    }
+  }, {
+    'underscored': true
+  });
+
+  // Each comment is made on an asset
+  Comment.belongsTo(Asset, {'onDelete': 'CASCADE'});
+
+  // Each comment is made by a user. We do not remove comments if a user is removed from the system
+  Comment.belongsTo(User);
```
nicolaasmatthijs: If we don't remove a comment when a user is removed from the system, how will we be able to display that comment?

node_modules/col-core/lib/db.js
```diff
@@ -78,19 +78,323 @@ var init = module.exports.init = function(callback) {
...
+  // but we will only use this for 1-level nesting
+  Comment.hasOne(Comment, {'as': 'parent'});
+
+  /**
+   * The `category` table keeps track of which categories assets can be tagged with
+   *
+   * @property  (String)      name      The name of the category
+   */
+  var Category = module.exports.Category = sequelize.define('category', {
+    'name': {
```
nicolaasmatthijs: Should we make this `displayname` as well?

nicolaasmatthijs:
Finished review. Back to @simong for follow-up

node_modules/col-core/lib/db.js
nicolaasmatthijs: I didn't realise that Canvas provides it as a comma-separated string. This makes sense.

node_modules/col-core/lib/db.js
simong: It doesn't look like that information is passed down in the LTI request

node_modules/col-core/lib/db.js
simong: You mentioned yesterday that comments should be retained, so I figured you mentioned something like this:
> *Removed user* posted:
> This is such a great asset

node_modules/col-core/lib/db.js
```diff
@@ -72,25 +72,332 @@ var init = module.exports.init = function(callback) {
...
+   * there's no overlap between courses, this is acceptable.
+   *
+   * @property  {Number}        canvas_user_id          The id of the user in Canvas
+   * @property  {String}        canvas_course_role      The role of the user in the course
+   * @property  {String}        canvas_full_name        The full name of the student
+   * @property  {String}        [canvas_image]          The URL that points to a profile picture for the user
+   * @property  {Number}        points                  The total Engagement Index points that the user has accumulated in the course
+   * @property  {Boolean}       share_points            Whether the user wants to share their point total with the other students in the course
+   */
+  var User = module.exports.User = sequelize.define('user', {
```
nicolaasmatthijs: This is missing the bookmarklet token

simong:
Followed-up, back to @nicolaasmatthijs

nicolaasmatthijs:
Followed up and merged

PR #7, opened 2015-03-13
COL-48 - Implement add link UI

nicolaasmatthijs:
- Implement add link UI
- Set up bookmarklet structure
- Basic asset list
- Track views on asset
- Track admin status for user

### DIFF
--- a/node_modules/col-assets/lib/rest.js
+++ b/node_modules/col-assets/lib/rest.js
@@ -21,7 +21,7 @@ var AssetsAPI = require('./api');
  * Get an asset
  */
 Collabosphere.apiRouter.get('/assets/:assetId', function(req, res) {
-  AssetsAPI.getAsset(req.ctx, req.params.assetId, function(err, asset) {
+  AssetsAPI.getAsset(req.ctx, req.params.assetId, req.query.increment_views, function(err, asset) {
     if (err) {
       return res.status(err.code).send(err.msg);
     }
--- a/node_modules/col-core/lib/constants.js
+++ b/node_modules/col-core/lib/constants.js
@@ -18,6 +18,7 @@ var Constants = module.exports = {
     'ACTIVITY_TYPES': ['like', 'dislike', 'comment', 'submit_assignment'],
     'OBJECT_TYPES': ['canvas_submission', 'canvas_discussion', 'asset', 'comment']
   },
+  'ADMIN_ROLES': ['Instructor', 'ContentDeveloper', 'urn:lti:role:ims/lis/TeachingAssistant'],
   'ASSET': {
     'ASSET_TYPES': ['file', 'link', 'whiteboard', 'thought']
   }
--- a/node_modules/col-core/lib/server.js
+++ b/node_modules/col-core/lib/server.js
@@ -17,6 +17,7 @@ var _ = require('lodash');
 var bodyParser = require('body-parser');
 var config = require('config');
 var cookieParser = require('cookie-parser');
+var cors = require('cors');
 var express = require('express');
 var http = require('http');
 var util = require('util');
@@ -52,6 +53,7 @@ var setUpServer = module.exports.setUpServer = function() {
    * Collabosphere supports the following type of request encodings:
    *
    *  * urlencoded (regular POST requests)
+   *  * application/json
    *  * TODO: multipart (file uploads)
    *
    * A maximum limit of 250kb is imposed for `urlencoded` and `application/json` requests. This limit only
@@ -59,6 +61,9 @@ var setUpServer = module.exports.setUpServer = function() {
    * using a proper multipart form request
    */
   app.use(bodyParser.urlencoded({'limit': '250kb', 'extended': true}));
+  app.use(bodyParser.json({'limit': '250kb'}));
+
+  app.use(cors());
 
   // Catch-all error handler
   app.use(function(err, req, res, next) {
--- a/node_modules/col-core/lib/server.js
+++ b/node_modules/col-core/lib/server.js
@@ -23,8 +23,6 @@ var util = require('util');
 
 var log = require('./logger')('col-core/server');
 
-var safePathPrefixes = [];
-
 /**
  * Start the Collabosphere Express server on the configured port
  *
@@ -62,41 +60,6 @@ var setUpServer = module.exports.setUpServer = function() {
    */
   app.use(bodyParser.urlencoded({'limit': '250kb', 'extended': true}));
 
-  /*!
-   * Referer-based CSRF protection. If the request is not safe (e.g., POST, DELETE) and the origin of the request (as
-   * specified by the HTTP Referer header) does not match the target host of the request (as specified by the HTTP
-   * Host header), then the request will result in a 500 error.
-   *
-   * While referer-based protection is not highly recommended due to spoofing possibilities in insecure environments,
-   * it currently offers the best trade-off between ease of use (e.g., for cURL interoperability), effort and security
-   * against CSRF attacks.
-   *
-   * Middleware that gets called earlier, can force the CSRF check to be skipped by setting `checkCSRF` on the request.
-   *
-   * If using a utility such as `curl` to POST requests to the API, you can bypass this by just setting the referer
-   * header to "/".
-   *
-   * More information about CSRF attacks: http://en.wikipedia.org/wiki/Cross-site_request_forgery
-   */
-  app.use(function(req, res, next) {
-    // If earlier middleware determined that CSRF is not required, we can skip the check
-    if (req.checkCSRF === false) {
-      return next();
-    }
-
-    if (!isSafeMethod(req.method) && !isSafePath(req) && !isSameOrigin(req)) {
-      log.warn({
-        'method': req.method,
-        'host': req.headers.host,
-        'referer': req.headers.referer,
-        'targetPath': req.path
-      }, 'CSRF validation failed: attempted to execute unsafe operation from untrusted origin');
-      return abort(res, 500, 'CSRF validation failed: attempted to execute unsafe method from untrusted origin');
-    }
-
-    return next();
-  });
-
   // Catch-all error handler
   app.use(function(err, req, res, next) {
     log.error({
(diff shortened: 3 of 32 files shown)


public/app/shared/utilService.js
```diff
@@ -0,0 +1,172 @@
...
+     * application
+     *
+     * @param  {Function}   messageGenerator              Function that will return the message to send to the parent container
+     * @api private
+     */
+    var postIFrameMessage = function(messageGenerator) {
+      // Only try to send the event when a parent container is present
+      if (window.parent) {
+        // Wait until Angular has finished rendering items on the screen
+        setTimeout(function() {
```
christianv: @nicolaasmatthijs could you try whether `$scope.$evalAsync()` works here? http://www.bennadel.com/blog/2605-scope-evalasync-vs-timeout-in-angularjs.htm

public/assets/css/main.css
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS"
+ * BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+body {
+  overflow-y: hidden;
```
christianv: @nicolaasmatthijs interesting, why did we need this?

public/bookmarklet.html
```diff
@@ -0,0 +1,112 @@
...
+        <div id="collabosphere-items-metadata" class="hide collabosphere-pane">
+          <h5>Add more information about the items you'd like to add</h5>
+          <ul id="collabosphere-items-metadata-container" class="list-unstyled collabosphere-items-metadata-container"><!-- --></ul>
+          <div class="modal-footer">
+            <button type="button" class="btn btn-link" id="collabosphere-items-metadata-back">Back</button>
+            <button type="button" class="btn btn-primary" id="collabosphere-items-add">Add</button>
+          </div>
+        </div>
+
+        <div id="collabosphere-done" class="hide collabosphere-pane"><!-- --></div>
```
christianv: @nicolaasmatthijs where is this used? Or is it just a placeholder for the future?

public/assets/css/main.css
nicolaasmatthijs: Without it, you can see the iFrame's scrollbar flicker in and out when the iFrame is being resized. Hiding the scrollbar essentially makes it a much nicer experience where the scrollbar flickering is invisible to the user. Data cultures uses this approach as well and it seems to be working well so far.

public/bookmarklet.html
nicolaasmatthijs: Correct. Just a placeholder for a future commit.

christianv:
Front-end lgtm

public/app/shared/utilService.js
nicolaasmatthijs: `$timeout` appears to work, but I can't get `$scope.$evalAsync` to work

nicolaasmatthijs:
Followed up on UI review in COL-50. As soon as this PR is merged, I should be able to keep future branches a bit more separated.

node_modules/col-core/lib/server.js
```diff
@@ -17,14 +17,13 @@ var _ = require('lodash');
 var bodyParser = require('body-parser');
 var config = require('config');
 var cookieParser = require('cookie-parser');
+var cors = require('cors');
```
simong: This needs to be added to package.json

simong: That said though, I haven't looked at what the module does, but if it's just adding headers this would be trivial with Apache

nicolaasmatthijs: Added it to `package.json` for now. I'll remove as part of the PR that introduce Apache and the production build. Just trying to keep things a bit staged for now.

node_modules/col-assets/lib/rest.js
```diff
@@ -21,7 +21,7 @@ var AssetsAPI = require('./api');
  * Get an asset
  */
 Collabosphere.apiRouter.get('/assets/:assetId', function(req, res) {
-  AssetsAPI.getAsset(req.ctx, req.params.assetId, function(err, asset) {
+  AssetsAPI.getAsset(req.ctx, req.params.assetId, req.query.increment_views, function(err, asset) {
```
simong: Should a REST call not always incur a "view increment" ?

node_modules/col-core/lib/constants.js
```diff
@@ -18,6 +18,7 @@ var Constants = module.exports = {
     'ACTIVITY_TYPES': ['like', 'dislike', 'comment', 'submit_assignment'],
     'OBJECT_TYPES': ['canvas_submission', 'canvas_discussion', 'asset', 'comment']
   },
+  'ADMIN_ROLES': ['Instructor', 'ContentDeveloper', 'urn:lti:role:ims/lis/TeachingAssistant'],
```
simong: What are the ETS people there roles within this course? One of these three?

node_modules/col-core/lib/server.js
```diff
@@ -61,49 +61,9 @@ var setUpServer = module.exports.setUpServer = function() {
...
-   * against CSRF attacks.
-   *
-   * Middleware that gets called earlier, can force the CSRF check to be skipped by setting `checkCSRF` on the request.
-   *
-   * If using a utility such as `curl` to POST requests to the API, you can bypass this by just setting the referer
-   * header to "/".
-   *
-   * More information about CSRF attacks: http://en.wikipedia.org/wiki/Cross-site_request_forgery
-   */
-  app.use(function(req, res, next) {
```
simong: I think we will still need the referer header check to void CSRF attacks. Maybe we can use something like https://www.npmjs.com/package/referer-filter ? We probably only need it on the `apiRouter`

node_modules/col-users/lib/rest.js
```diff
@@ -19,5 +19,12 @@ var Collabosphere = require('col-core');
...
  */
 Collabosphere.apiRouter.get('/me', function(req, res) {
-  return res.status(200).send(req.ctx.user);
+  if (!req.ctx) {
+    return res.status(401).send('Unauthenticated API request');
+  }
+
+  // Add the admin status to the response
+  var me = req.ctx.user.toJSON();
+  me.is_admin = req.ctx.user.isAdmin();
```
simong: I wonder if it makes sense to define a special getter for isAdmin (I believe they get included when serializing an object) or maybe override `User.toJSON`

node_modules/col-users/tests/util.js
```diff
@@ -13,31 +13,114 @@
...
+  assert.ok(me.id);
+  assert.ok(me.canvas_user_id);
+  assert.ok(me.course_id);
+  assert.ok(me.canvas_course_role);
+  assert.ok(me.canvas_full_name);
+  assert.ok(me.bookmarklet_token);
+  assert.ok(me.created_at);
+  assert.ok(me.updated_at);
+  assert(typeof me.points === 'number');
+  assert(typeof me.is_admin === 'boolean');
```
simong: isBoolean

node_modules/col-users/tests/util.js
```diff
@@ -13,31 +13,114 @@
...
+  assert.ok(me);
+  assert.ok(me.id);
+  assert.ok(me.canvas_user_id);
+  assert.ok(me.course_id);
+  assert.ok(me.canvas_course_role);
+  assert.ok(me.canvas_full_name);
+  assert.ok(me.bookmarklet_token);
+  assert.ok(me.created_at);
+  assert.ok(me.updated_at);
+  assert(typeof me.points === 'number');
```
simong: _.isFinite(me.points)

node_modules/col-users/tests/util.js
```diff
@@ -13,31 +13,114 @@
...
+  // Ensure that all expected properties are present on the course property
+  var course = me.course;
+  assert.ok(course);
+  assert.ok(course.id);
+  assert.strictEqual(course.id, me.course_id);
+  assert.ok(course.canvas_course_id);
+  assert.ok(course.canvas_api_domain);
+  assert.ok(course.created_at);
+  assert.ok(course.updated_at);
+  assert(typeof course.enable_upload === 'boolean');
```
simong: For these values as well

node_modules/col-users/tests/util.js
nicolaasmatthijs: This is taken care of as part of the COL-50 PR

node_modules/col-users/tests/util.js
nicolaasmatthijs: This is taken care of as part of the COL-50 PR

node_modules/col-users/tests/util.js
nicolaasmatthijs: This is taken care of as part of the COL-50 PR

node_modules/col-core/lib/constants.js
nicolaasmatthijs: This is the same admin roles list as used by Data Cultures

node_modules/col-assets/lib/rest.js
nicolaasmatthijs: I was thinking that the UI might want to request the asset without increasing the views (e.g. reloading the asset after making a comment). However, having done the REST APIs for comments, this is probably no longer necessary and we can just increase the views for every REST call.

node_modules/col-core/lib/server.js
nicolaasmatthijs: What about the API calls that the bookmarklet will be doing? We explicitly want those to be able to come from anywhere. Whitelisting those endpoints would open the door for CSRF attacks again. Does that mean that we should create dedicated endpoints for the bookmarklet APIs ?

node_modules/col-users/lib/rest.js
nicolaasmatthijs: Makes sense. I turned into a special getter.

node_modules/col-core/lib/server.js
simong: Can the bookmarklet not specify a `Referer` header? Alternatively, we could skip the CSRF check if `x-bookmarklet` authentication is used?

nicolaasmatthijs: jQuery doesn't allow a custom referrer to be specified. I'll try skipping the check if `x-bookmarklet` is specified. That probably means we won't be able to use `referer-filter` though.

PR #16, opened 2015-04-19
COL-83 - Upstart script

simong:
I added an upstart script that "daemonizes" collabosphere. It will keep respawning the node process when it crashes (up to 10 times per minute, it stops after that).

I also added an uncaughtException handler that will print the failure to the logs before exitting.

### DIFF
--- a/node_modules/col-core/lib/api.js
+++ b/node_modules/col-core/lib/api.js
@@ -34,6 +34,14 @@ var Server = require('./server');
  * @param  {Object}         callback.err        An error object, if any
  */
 var init = module.exports.init = function(callback) {
+  // All unexpected or uncaught errors will be caught and logged here. At this point we cannot
+  // guarantee that the system is functioning properly anymore so we kill the process. When running
+  // in production, the service script will automatically respawn the instance
+  process.on('uncaughtException', function(err) {
+    log.error({'err': err}, 'Uncaught exception was raised, restarting the process');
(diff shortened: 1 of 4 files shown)


nicolaasmatthijs:
Note that the test failures are caused by a 3rd party component upgrade and are fixed in one of my open PRs.

node_modules/col-core/lib/api.js
```diff
@@ -34,6 +34,14 @@ var Server = require('./server');
  * @param  {Object}         callback.err        An error object, if any
  */
 var init = module.exports.init = function(callback) {
+  // All unexpected or uncaught errors will be caught and logged here. At this point we cannot
+  // guarantee that the system is functioning properly anymore so we kill the process. When running
+  // in production, the service script will automatically respawn the instance
+  process.on('uncaughtException', function(err) {
```
nicolaasmatthijs: Do we have to exit the process here? In OAE, we did something like this:

```
// Set up application-level error handler
process.on('uncaughtException', function(err) {
    log().error({
        err : err
    }, 'An uncaught exception was raised to the application.');
});

// Handle the shutdown signal
process.on('SIGTERM', function() {
    Shutdown.shutdown(PRESHUTDOWN_DEFAULT_TIMEOUT_MILLIS, SHUTDOWN_GRACE_TIME_MILLIS);
});
```

PR #17, opened 2015-04-22
Do not use upstart script but rely on shell scripts and forever to daemonize collabosphere

simong:


deploy/K_collabosphere.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+#
+# The stop script for collabosphere. This needs to go into /home/app_collabosphere/init.d/
+
+# Change directory to where collabosphere should've been installed
+cd /home/app_collabosphere/collabosphere
```
nicolaasmatthijs: @simong : Can we do `~/...` instead of hardcoding `/home/app_collabosphere` ?

nicolaasmatthijs:
Looks good. Merged.

PR #18, opened 2015-05-01
COL-79: initial Selenium test for adding URL to asset library

pfarestveit:
https://jira.ets.berkeley.edu/jira/browse/COL-79

This is a first rspec test for adding URLs to the Asset Library.  The PR includes all the required test setup in Canvas (i.e., creating course site, adding test users and LTI tools), which will be reused in future tests.  The number of examples in the spec is minimal for now, though we should add tests for many permutations of URLs.

There are two issues outstanding:
* The YAML config file required by the WebDriverUtils
* A sample JSON file containing the test users data

For config, I defer to @nicolaasmatthijs.  Where to put the default file?  Should it be JSON instead?  And for the test data file, where should a sample live?

Finally, the code currently runs successfully on Firefox.  In order to play nicely with other browsers, it will require some future fine tuning.


selenium/logging.rb
```diff
@@ -0,0 +1,25 @@
+require 'logger'
```
nicolaasmatthijs: Can you give your files a license header ?

selenium/pages/asset_library_page.rb
```diff
@@ -0,0 +1,110 @@
+require_relative '../spec/spec_helper'
+
+class AssetLibraryPage
+
+  include PageObject
+  include Logging
+
+  # SEARCH
+  text_area(:search_input, :id => 'assetlibrary-list-search')
+  button(:advanced_search_button, :xpath => '//button[@title="Advanced search"]')
```
nicolaasmatthijs: Let me know if there are any elements that you'd like to see `id`s for

selenium/pages/asset_library_page.rb
```diff
@@ -0,0 +1,110 @@
...
+  elements(:category_title, :span, :xpath => '//span[@data-ng-bind="category.title"]')
+  elements(:edit_category_button, :button, :xpath => '//button[@title="Edit this category"]')
+  elements(:delete_category_button, :button, :xpath => '//button[@title="Delete this category"]')
+
+  # ASSETS
+  elements(:gallery_asset_link, :link, :xpath => '//li[@data-ng-repeat="asset in assets"]//a')
+  elements(:gallery_asset_title, :h3, :xpath => '//li[@data-ng-repeat="asset in assets"]//h3')
+  elements(:gallery_asset_owner_name, :element, :xpath => '//li[@data-ng-repeat="asset in assets"]//small')
+
+  def load_page(driver, url)
```
nicolaasmatthijs: I'd like us to maintain some basic documentation for each function (including documenting the parameters)

selenium/pages/asset_library_page.rb
```diff
@@ -0,0 +1,110 @@
...
+    navigate_to url
```
nicolaasmatthijs: Wouldn't it be better to keep map of page ids and page urls so not all function calls have to be adjusted when the page url is changed?

selenium/pages/asset_library_page.rb
```diff
@@ -0,0 +1,110 @@
...
+    wait_until(timeout=WebDriverUtils.page_load_wait) { driver.find_element(:id, 'tool_content') }
```
nicolaasmatthijs: Is this just a timeout after which it will check the existence of the element or will it wait for the element to become available? If there's a way to do the former approach, it would probably speed up the tests quite a bit.

selenium/pages/cal_net_page.rb
```diff
@@ -0,0 +1,20 @@
+require_relative '../spec/spec_helper'
+
+class CalNetPage
```
nicolaasmatthijs: Would be nice to see some documentation on what this is for

selenium/spec/asset_library_urls_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    @asset_library.delete_category(@driver, 1)
+  end
+
+  before(:example) do
+    @asset_library.load_page(@driver, @asset_library_url)
+    @asset_library.click_add_site_link
+  end
+
+  it 'allows the user to add a URL, title, category, and description to the asset library' do
+    @asset_library.enter_url_metadata('en.wikipedia.org', 'URL Title 1', 'Category 1', 'URL description')
```
nicolaasmatthijs: If this is an action that'll be taking place a lot, you might want to wrap these 2 functions into a "create_url" function that automatically generates some metadata and adds the link to the asset library

selenium/spec/asset_library_urls_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    @asset_library.click_add_url_button
+    @asset_library.wait_for_asset_in_gallery(@driver, 'URL Title 3')
+    expect(@asset_library.gallery_asset_owner_name_elements[0].text).to eql("by #{test_teacher['fullName']}")
+    @asset_library.click_asset_link 0
+    @asset_library.wait_for_asset_detail(@driver, 'URL Title 3')
+    # TODO: verify asset embedded URL once visible in the viewport
+  end
+
+  it 'allows the user to add a URL, title, and description to the asset library' do
+    @asset_library.enter_url_metadata('en.wikipedia.org', 'URL Title 4', nil, 'URL description')
```
nicolaasmatthijs: There's a lot of repetition between these functions, you might want to consolidate them

selenium/spec/asset_library_urls_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    @asset_library.missing_url_error_element.when_visible timeout=WebDriverUtils.page_update_wait
+    expect(@asset_library.url_title_input).to eql('URL Title 5')
+  end
+
+  it 'requires the user to enter a valid URL' do
+    @asset_library.enter_url_metadata('foo bar', 'URL Title 6', nil, nil)
+    @asset_library.click_add_url_button
+    @asset_library.bad_url_error_element.when_visible timeout=WebDriverUtils.page_update_wait
+    expect(@asset_library.url_input).to eql('http://foo bar')
+    expect(@asset_library.url_title_input).to eql('URL Title 6')
```
nicolaasmatthijs: Should this check the value of the validation message as well?

selenium/spec/asset_library_urls_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+  it 'does not offer a default category' do
+    expect(@asset_library.url_category).to eql('Which assignment or topic is this related to')
+  end
+
+  it 'offers only non-deleted categories' do
+    expect(@asset_library.url_category_options).to eql(['Which assignment or topic is this related to', 'Category 1'])
+  end
+
+  it 'limits a description to X characters' do
+    # TODO: verify max character validation if max is enforced
```
nicolaasmatthijs: The description doesn't have a max

selenium/util/web_driver_utils.rb
```diff
@@ -0,0 +1,77 @@
...
+  def self.engagement_index_url
+    @config['lti_engagement_index_url']
+  end
+
+  def self.test_course_name(spec)
+    "#{spec.inspect.sub('RSpec::ExampleGroups::', '')}-#{Time.now.to_i.to_s}"
+  end
+
+  def self.load_test_users
+    test_users = File.join(ENV['HOME'], '/.collabosphere_selenium/testUsers.json')
```
nicolaasmatthijs: Should this be included in the selenium folder or should we have some documentation on the format of this file?

nicolaasmatthijs:
Finished review. Back to @pfarestveit for follow-up

selenium/logging.rb
pfarestveit: @nicolaasmatthijs , yep will do.

selenium/pages/asset_library_page.rb
pfarestveit: The timeout is the max time it will wait for the iframe to show up, after which it will throw an error (and fail a test).  I believe it's checking for the iframe every half second, so I don't think there's much more speed to be achieved.

selenium/spec/asset_library_urls_spec.rb
pfarestveit: @nicolaasmatthijs from a design perspective, I like to be as explicit as possible about workflows in the tests themselves ... and therefore conservative about combining operations on the page.  But I agree that this is a pretty good candidate for consolidation.

selenium/spec/asset_library_urls_spec.rb
pfarestveit: @nicolaasmatthijs it actually is verifying the message, since the message element is defined by its text.

selenium/util/web_driver_utils.rb
pfarestveit: @nicolaasmatthijs we shouldn't put our actual test users file in the code, but we should definitely include an example file.  I'll add one to the folder.

selenium/spec/asset_library_urls_spec.rb
pfarestveit: @nicolaasmatthijs yeah, i agree.

selenium/pages/asset_library_page.rb
nicolaasmatthijs: Cool, that's how I was hoping it would work.

pfarestveit:
@nicolaasmatthijs - revisions are done.  I'd like to hold off on consolidating some of the repetitive method calls in the spec for now, if that's OK.  I'd prefer to do the consolidation when I add new tests for the asset library, since I think the best means of combining them will be clearer when I've got a good picture of all the usage.

nicolaasmatthijs:
Left some final review. Back to @pfarestveit for follow-up

pfarestveit:
Revisions done, back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #22, opened 2015-05-18
Canvas can return multiple roles over LTI

simong:
For example, somehow I ended up as: `Instructor,urn:lti:instrole:ims/lis/Administrator` on my local instance

### DIFF
--- a/node_modules/col-core/lib/db.js
+++ b/node_modules/col-core/lib/db.js
@@ -220,7 +220,8 @@ var setUpModel = function(sequelize) {
     'underscored': true,
     'getterMethods': {
       'is_admin': function() {
-        return _.contains(CollabosphereConstants.ADMIN_ROLES, this.canvas_course_role);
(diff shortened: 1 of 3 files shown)


node_modules/col-core/lib/db.js
```diff
@@ -220,7 +220,8 @@ var setUpModel = function(sequelize) {
     'underscored': true,
     'getterMethods': {
       'is_admin': function() {
-        return _.contains(CollabosphereConstants.ADMIN_ROLES, this.canvas_course_role);
+        var roles = this.canvas_course_role.split(',');
+        return (_.intersection(CollabosphereConstants.ADMIN_ROLES, roles).length !== 0);
```
nicolaasmatthijs: I'm going to make this `> 0` just to be a bit more specific

nicolaasmatthijs:
Followed up and merged

PR #21, opened 2015-05-17
Engagement index, asset library refinements and base whiteboard functionality

nicolaasmatthijs:
- Engagement index CSV export
- Asset library likes
- Asset library reply, edit and delete comments
- Convert links in comments
- Manage categories
- Engagement index UI
- COL-95 - Asset Library: gallery view broken up by whitespace in Chrome
- COL-92 - Asset Library: should validate title length
- Base whiteboard functionality

PR #24, opened 2015-05-22
Revert "Engagement index, asset library refinements and base whiteboa…

simong:
Reverts ets-berkeley-edu/collabosphere#21


simong:
I merged the wrong PR before :-1:

PR #26, opened 2015-05-22
Engagement index, asset library refinements and base whiteboard functionality

nicolaasmatthijs:
- Engagement index CSV export
- Asset library likes
- Asset library reply, edit and delete comments
- Convert links in comments
- Manage categories
- Engagement index UI
- COL-95 - Asset Library: gallery view broken up by whitespace in Chrome
- COL-92 - Asset Library: should validate title length
- Base whiteboard functionality


nicolaasmatthijs:
Rebased against master

apache/collabosphere.template
```diff
@@ -14,12 +14,22 @@
     CustomLog <%= logDirectory %>/collabosphere_access_log common
     CustomLog <%= logDirectory %>/collabosphere-combined_log combined
 
+    RewriteEngine on
+    RewriteCond %{QUERY_STRING} transport=polling
+    RewriteRule /(.*)$ http://localhost:2000/$1 [P]
```
simong: Can you add some documentation that this rewrite rule is forcing the request to be proxied to the node app?
Additionally, can you add why we can't add another ProxyPass entry?

simong: Err, I must be getting blind, there's a ProxyPass entry 2 lines lower

node_modules/col-core/lib/api.js
```diff
@@ -34,12 +34,9 @@ var Server = require('./server');
  * @param  {Object}         callback.err        An error object, if any
  */
 var init = module.exports.init = function(callback) {
-  // All unexpected or uncaught errors will be caught and logged here. At this point we cannot
-  // guarantee that the system is functioning properly anymore so we kill the process. When running
-  // in production, the service script will automatically respawn the instance
+  // All unexpected or uncaught errors will be caught and logged here
   process.on('uncaughtException', function(err) {
-    log.error({'err': err}, 'Uncaught exception was raised, restarting the process');
-    process.exit(1);
```
simong: Why was this removed and how will the service recover?

node_modules/col-core/lib/db.js
```diff
@@ -407,6 +407,126 @@ var setUpModel = function(sequelize) {
   Asset.belongsToMany(Category, {'through': 'assets_categories', 'onDelete': 'CASCADE'});
 
   /**
+   * The `whiteboards` table will keep track of the available whiteboards in the Whiteboard tool
+   *
+   * @property {String}       title                   The title of the whiteboard
+   */
+  var Whiteboard = module.exports.Whiteboard = sequelize.define('whiteboard', {
```
simong: Do we need to keep track of who created the whiteboard?

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+    } else if (!whiteboard) {
+      log.debug({'err': err, 'id': id}, 'A whiteboard with the specified id could not be found');
+      return callback({'code': 404, 'msg': 'A whiteboard with the specified id could not be found'});
+    }
+
+    whiteboard = whiteboard.toJSON();
+
+    // Parse the stringified whiteboard elements and return
+    // them in a format that can be understood by Fabric.js
+    // TODO: Look at `_.map` for this instead
```
simong: Something like this maybe?

```javascript
try {
    whiteboard.whiteboard_elements = _.map(whiteboard.whiteboard_elements, function(whiteboardElement) {
        return JSON.parse(whiteboardElement.element);
    });
} catch (err) {
    log.error({'err': err, 'whiteboard': id}, 'A whiteboard contained malformed JSON');
}
```

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+ * Get a whiteboard
+ * // TODO: Only load online people and whiteboard content when required
+ *
+ * @param  {Number}         userId                          The id of the user retrieving the whiteboard
+ * @param  {Number}         id                              The id of the whiteboard
+ * @param  {Function}       callback                        Standard callback function
+ * @param  {Object}         callback.err                    An error that occurred, if any
+ * @param  {Whiteboard}     callback.whiteboard             The requested whiteboard
+ */
+var getWhiteboard = module.exports.getWhiteboard = function(userId, id, callback) {
```
simong: We should pass in the request context here rather than the userId. That will allow us to add the course to the `where` parameters. As it stands, users from other courses could potentially request whiteboards they shouldn't be able to see

simong: That's assuming all users from a course can see all the whiteboards created within that course?

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+    if (err) {
+      log.error({'err': err}, 'Failed to delete all whiteboard session');
+      return callback({'code': 500, 'msg': err.message});
+    }
+
+    return callback();
+  });
+};
+
+// Delete all whiteboard sessions when the server starts
```
simong: How would this work in a situation where the number of appservers is > 1 ? Is that something that should be taken into account? Will whiteboards have some affinity to an app server? etc..

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+deleteWhiteboardSessions();
+
+/* WHITEBOARD ELEMENTS */
+
+/**
+ * Store a new element was added to the whiteboard canvas and inform
```
simong: `Store a new element that was added`

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+      log.debug({'socket': socket.id, 'element': element.uid}, 'An element on the whiteboard was deleted');
+      deleteWhiteboardElement(socket, element);
+    });
+
+    /**
+     * When a user disconnected, they are removed from the room and their online record
+     * is deleted as well
+     */
+    socket.on('disconnect', function() {
+      log.info({'socket': socket.id}, 'Websocket connection was disconnected');
```
simong: This can probably be demoted to `debug`

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+ * whiteboard from which they connected. A record indicating that the user is online in
+ * the whiteboard will also be added
+ */
+Collabosphere.appServer.io.on('connection', function(socket) {
+  log.info({'socket': socket.id}, 'Initiating new websocket connection');
+
+  // Extract the API domain and course id from the referer URL, as well
+  // as the room to which the user has connected. The expected referer URL
+  // format is `http://domain/whiteboards/<whiteboard_id>?api_domain=<api_domain>&course_id=<course_id>`
+  var referer = url.parse(socket.handshake.headers.referer, true);
```
simong: Is there no way to pass the api domain / course id in the handshake itself?

node_modules/col-core/lib/api.js
nicolaasmatthijs: I removed it because this would cause all of the connected websockets to be disconnected. They do reconnect after a while, but there'll be a time gap where those users won't be able to make whiteboard changes. Let me know if you would like me to undo this change.

node_modules/col-core/lib/db.js
nicolaasmatthijs: This will be implemented as part of access management for whiteboards, which will be part of a follow-up PR.

node_modules/col-whiteboards/lib/api.js
nicolaasmatthijs: I've outscoped multi-appserver support from this iteration, as the service will probably be run like that. However, if you think it's important to give it some consideration or if you think it'll be hard to retrofit further down the line, we can give it some thought now.

node_modules/col-whiteboards/lib/api.js
nicolaasmatthijs: The problem is that we need to use `getWhiteboard` in some of the socket code, where we don't have access to a Context object. Whiteboards will become access controlled, but lecturers should be able to see all whiteboards within their course, so the course id will need to be available here.

Do you want me to just pass in the courseId as well or should I create a `getWhiteboard` and `_ getWhiteboard` function, one that takes a context and the internal one that takes `userId` and `courseId`?

node_modules/col-core/lib/api.js
simong: If an error bubbles up to this global catch-all handler the application is pretty much in an unknown state. The recommended practice is to restart your application and start from a clean slate: https://nodejs.org/api/process.html#process_event_uncaughtexception

> Note that uncaughtException is a very crude mechanism for exception handling.
> Do not use it as the node.js equivalent of On Error Resume Next. An unhandled exception means your application - and by extension node.js itself - is in an undefined state. Blindly resuming means anything could happen.
> Think of resuming as pulling the power cord when you are upgrading your system. Nine out of ten times nothing happens - but the 10th time, your system is bust.
> 
> You have been warned.

We should strive to never hit this block of logic, when we do, we should probably restart gracefully.

node_modules/col-whiteboards/lib/api.js
nicolaasmatthijs: Found a way to pass this in a bit more cleanly

node_modules/col-whiteboards/lib/api.js
simong: When a new socket connection comes in through `Collabosphere.appServer.io.on('connection'` can we not do something like:
```javascript
Collabosphere.appServer.io.on('connection', function(socket) {
  // Cookie stuff
  var userId = cookieParser.signedCookie(sessionCookie, config.get('cookie.secret'));

  // Referer stuff
  var whiteboardId = referer.pathname.split('/').pop();

  // Get the user
  UsersAPI.getUser(userId, function(err, user) {
    if (err) { return callback(err); }
    // Keep track of the context
    socket.ctx = {'user': user, 'course': user.course};

    // Get whiteboard
    getWhiteboard(socket.ctx, whiteboardId, function(err, whiteboard) {
       ....
    });
  });
};
```
get the user object from which we can generate a context? That would allow us to keep our usual style of passing in a ctx as the first parameter of each API method.

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+ * @param  {Function}       [callback]                      Standard callback function
+ * @param  {Object}         [callback.err]                  An error that occurred, if any
+ * @api private
+ */
+var addWhiteboardElement = function(socket, element, callback) {
+  callback = callback || function() {};
+
+  // Notify all online users in the whiteboard that a new whiteboard element was added.
+  // This will not send a notification over the socket on which the new element was sent
+  socket.broadcast.to(socket.whiteboard.id).emit('addElement', element);
```
simong: Should this broadcast whappen when `upsertWhiteboardElement`'s callback function got triggered?

node_modules/col-whiteboards/lib/api.js
```diff
@@ -0,0 +1,771 @@
...
+ * @param  {Function}       [callback]                      Standard callback function
+ * @param  {Object}         [callback.err]                  An error that occurred, if any
+ * @api private
+ */
+var updateWhiteboardElement = function(socket, element, callback) {
+  callback = callback || function() {};
+
+  // Notify all online users in the whiteboard that a whiteboard element was updated.
+  // This will not send a notification over the socket on which the element update was sent
+  socket.broadcast.to(socket.whiteboard.id).emit('updateElement', element);
```
simong: Should this broadcast whappen when upsertWhiteboardElement's callback function got triggered?

public/app/shared/utilService.js
```diff
@@ -27,6 +27,21 @@
...
+     *
+     * @return {Object}                       The parameters that were passed in through the iFrame launch URL
+     */
+    var getLaunchParams = function() {
+      var launchParams = {
+        'apiDomain': apiDomain,
+        'courseId': courseId,
+        'toolUrl': toolUrl
+      };
+      console.log(launchParams);
```
simong: `console.log`

public/app/app.config.js
```diff
@@ -17,7 +17,9 @@
 
   'use strict';
 
-  angular.module('collabosphere').config(function($compileProvider) {
+  angular.module('collabosphere').config(function($compileProvider, $tooltipProvider) {
+
+    console.log($tooltipProvider);
```
simong: `console.log`

public/app/whiteboards/create/create.html
```diff
@@ -0,0 +1,35 @@
...
+        <input type="text" id="whiteboards-create-title" name="whiteboardsCreateTitle" class="form-control" placeholder="Enter a title" data-ng-model="whiteboard.title" data-ng-maxlength="255" required>
+        <div class="help-block" data-ng-messages="whiteboardsCreateForm.whiteboardsCreateTitle.$error">
+          <div data-ng-message="required">Please enter a title</div>
+          <div data-ng-message="maxlength">A title can only be 255 characters long</div>
+        </div>
+      </div>
+    </div>
+
+    <!-- INVITE -->
+    <div class="form-group" data-ng-class="{'has-error': whiteboardsCreateForm.$submitted && whiteboardsCreateForm.whiteboardsCreateInvite.$invalid}">
```
simong: I assume this is separate from the user listing in the engagement index? e.g., you'll be able to search through the full set of users regardless of whether they (or yourself) have accepted to share in the engagement index.

public/app/whiteboards/create/create.html
```diff
@@ -0,0 +1,35 @@
...
+      <div class="col-sm-10">
+        <input type="text" id="whiteboards-create-invite" name="whiteboardsCreateInvite" class="form-control" placeholder="Invite students into this whiteboard">
+        <div class="help-block" data-ng-messages="whiteboardsCreateForm.whiteboardsCreateInvite.$error"><!-- --></div>
+      </div>
+    </div>
+
+    <!-- BUTTONS -->
+    <div class="form-group text-right">
+      <div class="col-sm-offset-2 col-sm-10">
+        <a data-ng-href="/whiteboards" class="btn btn-default">Cancel</a>
```
simong: This button currently doesn't do anything

public/app/whiteboards/create/create.html
nicolaasmatthijs: Correct. There'll need to be a separate feed for listing all users regardless of their points share state.

public/app/whiteboards/create/create.html
nicolaasmatthijs: It links back to the whiteboard overview page

node_modules/col-whiteboards/lib/api.js
nicolaasmatthijs: The thinking was that we want to make the update available to the other users in the whiteboard as soon as possible. However, given that this should only be a matter of ms, I'll change it to broadcast in the callback.

nicolaasmatthijs:
Followed up on review. Back to @simong for review.

simong:
Neat! Merging this in

PR #97, opened 2015-08-13
COL-296 - Support picture tags in the bookmarklet

simong:


paulkerschen:
Closing per discussion above.

PR #113, opened 2015-09-14
COL-333 - Store a whiteboard’s full PNG image when exporting to the asset library

simong:


paulkerschen:
Looks good to me @simong; thanks.

PR #119, opened 2016-01-15
COL-369 - Basic migration script for file and link assets

nicolaasmatthijs:
Basic migration script for file and link assets

https://jira.ets.berkeley.edu/jira/browse/COL-369


nicolaasmatthijs:

node_modules/col-core/lib/db.js
```diff
@@ -390,7 +393,7 @@ var setUpModel = function(sequelize) {
 
         // Add the liked status
         asset.liked = null;
-        if (asset.activities.length > 0) {
+        if (asset.activities && asset.activities.length > 0) {
```
simong: you can just do `!_.isEmpty(asset.activities)) {`

scripts/migrate_assets.js
```diff
@@ -0,0 +1,229 @@
...
+      }, {
+        'description': file.description || undefined,
+        'thumbnail_url': file.thumbnail_url || undefined,
+        'image_url': file.image_url || undefined,
+        'embed_id': file.embed_id || undefined,
+        'embed_key': file.embed_key || undefined,
+        'embed_code': file.embed_code || undefined
+      }, function(err, asset) {
+        if (err) {
+          return callback(err);
```
simong: The `temp` file needs to be cleaned up here

scripts/migrate_assets.js
```diff
@@ -0,0 +1,229 @@
...
+    .on('error', callback)
+    .on('finish', function() {
+
+      // Create the file asset
+      AssetsAPI.createFile(ctx, file.title, {
+        'mimetype': file.mime,
+        'file': path,
+        'filename': file.title
+      }, {
+        'description': file.description || undefined,
```
simong: Why are the `|| undefined`  here?

scripts/migrate_assets.js
```diff
@@ -0,0 +1,229 @@
...
+ * Migrate a file asset
+ *
+ * @param  {Asset}            file                The file to migrate
+ * @param  {Function}         callback            Standard callback function
+ */
+var migrateFile = function(file, callback) {
+  // Download the file to a temporary folder
+  var path = temp.path();
+  request(file.download_url).pipe(fs.createWriteStream(path))
+    .on('error', callback)
```
simong: The `temp` file needs to be deleted here as well (if there is any). It might be easier to wrap `callback` and always try to clean up the file

scripts/migrate_assets.js
```diff
@@ -0,0 +1,229 @@
...
+
+    log.info('Connected to the database');
+
+    // Create the context for interaction with the API
+    createContext();
+  });
+};
+
+/**
+ * Create a mock context to use for interaction with the API
```
simong: This function does more than just create a context. It also kicks off the entire migration process

simong:
Assuming we only want to migrate content for 1 user (John Scott), what does the workflow look like here?

1. Create the new course
2. Add the tools to the course
3. Run the poller so John Scott gets a User account
4. Run the migration script

What about the links in the Pages/wiki? Is that a todo or a manual step?

scripts/migrate_assets.js
nicolaasmatthijs: We're calling the internal API directly and Joi throws an error for optional values when `null` is provided. I didn't really want to add `allow(null)` to all validation rules, so I just default them to `undefined` for all falsey values. However, I'm happy to go for an alternative solution here. What do you think?

nicolaasmatthijs:
> Assuming we only want to migrate content for 1 user (John Scott), what does the workflow look like here?

Yes, the migration workflow is the one you describe:

1. Create the new course and add the user to it
2. Add the tools
3. Ensure that the course and the user have an entry in the Collabosphere database
4. Run the script

I've agreed with John that updating the links in the Modules and Page is a manual step for now.

nicolaasmatthijs:
Followed up on review and added category migration

nicolaasmatthijs:
Incorporated shrinkwrap and some additional fixes to make the migration script run in production Canvas

nicolaasmatthijs:
This is ready for more review.

node_modules/col-assets/tests/test-assets.js
```diff
@@ -68,7 +68,12 @@ describe('Assets', function() {
               // Create a link asset with optional metadata
               var opts = {
                 'description': 'University of California, Berkeley homepage',
-                'source': 'http://www.universityofcalifornia.edu/uc-system'
+                'source': 'http://www.universityofcalifornia.edu/uc-system',
```
simong: Can you assert that all these values were not set through the REST API ?

nicolaasmatthijs:
Followed up. Back to @simong for review

PR #120, opened 2016-01-19
Shrinkwrap all dependencies

simong:
Production isn't failing on `npm install` because of the following dependency chain:
fabric -> jsdom -> contextify@0.1.15 -> nan@2.1.0

nan 2+ assumes that you're running a node binary with a patched v8 library. This is true for all official node binaries, but is most likely not the case for packages distributed by distros such as Debian, Ubuntu, .. as they would build against a shared v8 library. As we don't really need any of the updates yet, I've provided a shrinkwrap that fixes contextify to 0.1.4 (and nan to 1.8.4).


nicolaasmatthijs:
Looks good. Merged.

PR #122, opened 2016-01-23
Use hashed images when sending emails

simong:


config/default.json
```diff
@@ -4,7 +4,7 @@
     "apiKey": "ed26bc504e17992d16554c70f8bb3914"
   },
   "apache": {
-    "documentRoot": "/var/www/html",
+    "documentRoot": "./target",
```
nicolaasmatthijs: I'm going to change this back

node_modules/col-core/lib/email.js
```diff
@@ -52,6 +52,26 @@ var EVENT_NAMES = module.exports.EVENT_NAMES = {
   'EMAIL_SENT': 'emailSent'
 };
 
+// Variable that maps the original image paths to their hashed counter parts (if any)
+var imageManifest = {};
+
+// Cache the image manifest
+var manifestPath = path.resolve(__dirname, '..', '..', '..', config.get('apache.documentRoot'), 'images-rev-manifest.json');
```
nicolaasmatthijs: Isn't it easier to just keep `documentRoot` the way it was as it points to where the static files are? If it's a production build, `images-rev-manifest.json` should be there. It would avoid having to hardcode going up 3 levels.

node_modules/col-core/lib/email.js
```diff
@@ -52,6 +52,26 @@ var EVENT_NAMES = module.exports.EVENT_NAMES = {
...
+if (fs.existsSync(manifestPath)) {
+  try {
+    var data = fs.readFileSync(manifestPath).toString();
```
nicolaasmatthijs: I think you can just do a require on that file ?

node_modules/col-core/lib/email.js
```diff
@@ -101,6 +121,21 @@ var sendEmail = module.exports.sendEmail = function(subject, user, course, data,
     },
 
     /**
+     * Resolve the hashed URL for a static asset
+     *
+     * @param  {String} assetPath The path to the asset (including leading /)
```
nicolaasmatthijs: Spacing around `assetPath`

nicolaasmatthijs:
Finished review. Back to @simong for follow-up

node_modules/col-core/lib/email.js
simong: Yes, that's what `documentRoot` is supposed to do. This is supposed to be `path.resolve(config.get('apache.documentRoot'), 'images-rev-manifest.json')` but I forgot to clean up my debugging code :(

nicolaasmatthijs:
Looks good. Merged

PR #123, opened 2016-01-24
Trigger a popup if a whiteboard is deeplinked

simong:


public/app/assetlibrary/assetLibraryService.js
```diff
@@ -32,24 +32,7 @@
     // Get the parent container's hash. In case a hash a present, restore the state to allow
```
nicolaasmatthijs: Typo + this is now no longer 100% accurate

public/app/shared/utilService.js
```diff
@@ -166,20 +166,69 @@
...
           'subject': 'getParent'
         };
       }, function(data) {
-        data = data || {};
-        return callback(data.location);
+        if (!data || !data.location) {
+          return callback({});
+        }
+
+        // Parse the URL
```
nicolaasmatthijs: I'm going to check if we can use a proper library for this

nicolaasmatthijs:
Followed up and merged

PR #124, opened 2016-03-15
COL-376 - Tmp space fills up

nicolaasmatthijs:
https://jira.ets.berkeley.edu/jira/browse/COL-376


nicolaasmatthijs:

PR #126, opened 2016-05-18
COL-340 When user removes self from whiteboard, close and refresh list

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-340

One UX question: this does make it easier for a user to inadvertently lose access to a whiteboard. Is it worth a confirmation dialog?

### DIFF
--- a/public/app/whiteboards/board/whiteboardsBoardDirective.js
+++ b/public/app/whiteboards/board/whiteboardsBoardDirective.js
@@ -48,7 +48,7 @@
         'readonly': '=readonly'
       },
       'templateUrl': '/app/whiteboards/board/board.html',
-      'controller': function(analyticsService, Fabric, FabricConstants, me, utilService, whiteboardsFactory, $alert, $cookies, $modal, $rootScope, $scope) {
+      'controller': function(analyticsService, Fabric, FabricConstants, me, utilService, whiteboardsFactory, $alert, $cookies, $modal, $rootScope, $scope, $window) {
 
         // Make the me object available to the scope
         $scope.me = me;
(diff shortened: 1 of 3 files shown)


nicolaasmatthijs:
@paulkerschen : With regards to the UX question, we haven't necessarily made it easier to lose access to a whiteboard. While removing yourself wouldn't have closed the window previously, you still would have lost the ability to add yourself back in. Having said that, I do think adding a confirmation dialog is a good idea here.

public/app/whiteboards/board/whiteboardsBoardDirective.js
```diff
@@ -1794,9 +1794,15 @@
           scope.whiteboard = $scope.whiteboard;
           scope.closeModal = function(updatedWhiteboard) {
             if (updatedWhiteboard) {
-              $scope.whiteboard = updatedWhiteboard;
-              // Set the title of the window to the new title of the whiteboard
-              $rootScope.header = $scope.whiteboard.title;
+              if (updatedWhiteboard.notFound) {
+                // If an edit has removed the user's access, close and refresh the main whiteboard list
+                $window.close();
+                $window.opener.refreshWhiteboardList();
```
nicolaasmatthijs: Could there be a potential issue by using the opener after we've already called `close` on the window?

nicolaasmatthijs:
Left one minor code comment and some feedback on the proposed UX change. Back to @paulkerschen for follow-up

public/app/whiteboards/board/whiteboardsBoardDirective.js
paulkerschen: Could be, thanks, @nicolaasmatthijs. Reordered, and added a null check on the opener in case it's been closed in the meantime.

paulkerschen:
Right, we haven't made self-removal easier, we're just explicitly contemplating that path now. Confirmation dialog added and code comment addressed, back to @nicolaasmatthijs for review.

paulkerschen:
Travis seems to be erroring out on https://github.com/npm/npm/issues/12001. I got things to build locally by manually changing npm-shrinkwrap.json from which-1.2.2 to which-1.2.4, but I'm too much a newbie to know what's the cleanest fix.

nicolaasmatthijs:
@paulkerschen : Would you mind keeping code review follow-up commits separate from the original commit (i.e., not squash them)? This makes it easier to just review the code review follow-up changes.

nicolaasmatthijs:
Looks good. Merged.

PR #131, opened 2016-05-19
COL-334 Fix background image retrieval in bookmarklet

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-334


nicolaasmatthijs:
Good catch. Merged.

PR #133, opened 2016-05-20
COL-352 Replace form submit in asset search with event listeners

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-352

Fixes unresponsive submit when the asset search form is nested inside another form in a modal.

### DIFF
--- a/public/app/assetlibrary/search/assetLibrarySearchDirective.js
+++ b/public/app/assetlibrary/search/assetLibrarySearchDirective.js
@@ -87,6 +87,15 @@
         };
 
         /**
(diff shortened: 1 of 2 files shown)


nicolaasmatthijs:
In thinking about this a bit more, I was wondering if we can just wrap the form around the `Cancel` and `Add selected` buttons, so that we have a form for both the search action and the submit action.

nicolaasmatthijs:
Back to @paulkerschen

paulkerschen:
A good suggestion, @nicolaasmatthijs: simpler and appears to work. Closing this PR in favor of #136.

PR #134, opened 2016-05-23
COL-378 Allow non-admin users to delete associated assets with no interactions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-378

### DIFF
--- a/node_modules/col-assets/lib/api.js
+++ b/node_modules/col-assets/lib/api.js
@@ -55,7 +55,8 @@ var UserConstants = require('col-users/lib/constants');
 var getAssetProfile = module.exports.getAssetProfile = function(ctx, id, incrementViews, callback) {
   var opts = {
     'incrementViews': incrementViews,
-    'includeWhiteboardElements': true
+    'includeWhiteboardElements': true,
+    'includeWhiteboardUsages': true
   };
   getAsset(ctx, id, opts, function(err, asset) {
     if (err) {
@@ -76,6 +77,7 @@ var getAssetProfile = module.exports.getAssetProfile = function(ctx, id, increme
  * @param  {Object}         [opts]                                  A set of options that determine what data will be retrieved
  * @param  {Boolean}        [opts.incrementViews]                   Whether the total number of views for the asset should be incremented by 1
  * @param  {Boolean}        [opts.includeWhiteboardElements]        Whether the asset whiteboard elements should be included
+ * @param  {Boolean}        [opts.includeWhiteboardUsages]          Whether the asset's whiteboard usages should be included
  * @param  {Function}       callback                                Standard callback function
  * @param  {Object}         callback.err                            An error that occurred, if any
  * @param  {Asset}          callback.asset                          The requested asset
@@ -88,7 +90,8 @@ var getAsset = module.exports.getAsset = function(ctx, id, opts, callback) {
     'id': Joi.number().required(),
     'opts': Joi.object().keys({
       'incrementViews': Joi.boolean().optional(),
-      'includeWhiteboardElements': Joi.boolean().optional()
+      'includeWhiteboardElements': Joi.boolean().optional(),
+      'includeWhiteboardUsages': Joi.boolean().optional(),
     })
   });
 
@@ -141,6 +144,14 @@ var getAsset = module.exports.getAsset = function(ctx, id, opts, callback) {
     options.include.push({
       'model': DB.AssetWhiteboardElement,
       'as': 'whiteboard_elements',
+      'required': false
+    });
+  }
+  if (opts.includeWhiteboardUsages) {
+    options.include.push({
+      'model': DB.WhiteboardElement,
+      'attributes': ['created_at', 'whiteboard_id'],
+      'as': 'whiteboard_usages',
       'required': false
     });
   }
@@ -824,16 +835,24 @@ var deleteAsset = module.exports.deleteAsset = function(ctx, id, callback) {
     return callback({'code': 400, 'msg': validationResult.error.details[0].message});
   }
 
-  // Only course instructors are allowed to delete an asset
-  if (!ctx.user.is_admin) {
-    log.error({'id': id}, 'Unauthorized to delete an asset');
-    return callback({'code': 401, 'msg': 'Unauthorized to delete an asset'});
-  }
-
   // Retrieve the asset that is being deleted
-  getAsset(ctx, id, null, function(err, asset) {
+  var opts = {
+    'includeWhiteboardUsages': true
+  };
+  getAsset(ctx, id, opts, function(err, asset) {
     if (err) {
       return callback(err);
+    }
+
+    // If an asset has no interactions, associated users may delete that asset. Other assets may be deleted
+    // only by course instructors
+    if (!ctx.user.is_admin) {
+      var isUser = (_.find(asset.users, {'id': ctx.user.id}));
+      var hasNoInteractions = !(asset.comment_count || asset.dislikes || asset.likes) && _.isEmpty(asset.whiteboard_usages);
+      if (!(isUser && hasNoInteractions)) {
+        log.error({'id': id}, 'Unauthorized to delete an asset');
+        return callback({'code': 401, 'msg': 'Unauthorized to delete an asset'});
+      }
     }
 
     // Delete the asset from the DB
--- a/node_modules/col-assets/lib/api.js
+++ b/node_modules/col-assets/lib/api.js
@@ -42,8 +42,10 @@ var log = require('col-core/lib/logger')('col-assets');
 var UserConstants = require('col-users/lib/constants');
 
 /**
- * Get a full asset profile. Next to the basic asset profile, this will return the whiteboard
- * elements in a form that Fabric.js can understand
+ * Get a full asset profile. Next to the basic asset profile, this will return:
+ *  - the whiteboard elements in a form that Fabric.js can understand;
+ *  - any exported whiteboards that use this asset;
+ *  - a flag indicating whether the current user is able to delete this asset.
  *
  * @param  {Context}        ctx                     Standard context containing the current user and the current course
  * @param  {Number}         id                      The id of the asset
@@ -65,6 +67,13 @@ var getAssetProfile = module.exports.getAssetProfile = function(ctx, id, increme
 
     asset = asset.toJSON();
     asset.whiteboard_elements = _.pluck(asset.whiteboard_elements, ['element']);
+    asset.exported_whiteboards = _.pluck(asset.exported_whiteboards, ['id', 'title']);
+
+    // Usage in active whiteboards will prevent an asset from being deleted, but should not be kept in the asset profile
+    // once the canDelete property is set
+    asset.canDelete = canDeleteAsset(ctx.user, asset);
+    delete asset.whiteboard_usages;
+
     return callback(null, asset);
   });
 };
@@ -150,8 +159,14 @@ var getAsset = module.exports.getAsset = function(ctx, id, opts, callback) {
   if (opts.includeWhiteboardUsages) {
     options.include.push({
       'model': DB.WhiteboardElement,
-      'attributes': ['created_at', 'whiteboard_id'],
+      'attributes': ['whiteboard_id'],
       'as': 'whiteboard_usages',
+      'required': false
+    });
+    options.include.push({
+      'model': DB.Asset,
+      'attributes': ['title', 'id'],
+      'as': 'exported_whiteboards',
       'required': false
     });
   }
@@ -625,7 +640,8 @@ var createWhiteboard = module.exports.createWhiteboard = function(ctx, whiteboar
         // Prepend the current date to the `uid` as the same board can be exported multiple times
         'uid': util.format('%d-%s', Date.now(), element.uid),
         'element': element.element,
-        'asset_id': asset.id
+        'asset_id': asset.id,
+        'element_asset_id': element.asset_id
       };
     });
     DB.AssetWhiteboardElement.bulkCreate(bulkWhiteboardelements).complete(function(err) {
@@ -773,7 +789,7 @@ var editAsset = module.exports.editAsset = function(ctx, id, title, opts, callba
       return callback(err);
     }
 
-    // Only course instructors and the user that has created the asset are allowed
+    // Only course administrators and the user that has created the asset are allowed
     // to edit it
     if (!ctx.user.is_admin && !_.find(asset.users, {'id': ctx.user.id})) {
       log.error({'course': ctx.course}, 'Unauthorized to edit an asset');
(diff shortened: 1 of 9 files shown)


node_modules/col-assets/lib/api.js
```diff
@@ -55,7 +55,8 @@ var UserConstants = require('col-users/lib/constants');
 var getAssetProfile = module.exports.getAssetProfile = function(ctx, id, incrementViews, callback) {
   var opts = {
     'incrementViews': incrementViews,
-    'includeWhiteboardElements': true
+    'includeWhiteboardElements': true,
+    'includeWhiteboardUsages': true
   };
```
nicolaasmatthijs: Rather than including the whiteboard elements in which an asset is being used, can we determine whether the user can delete the asset in the back-end and just return a `canDelete` property as part of the asset profile?

paulkerschen: I almost went that way, @nicolaasmatthijs, but thought that the front end might want to associate assets with whiteboards for other reasons down the line. We can move it all to the back end if you'd prefer, though.

node_modules/col-assets/lib/api.js
```diff
@@ -144,6 +147,14 @@ var getAsset = module.exports.getAsset = function(ctx, id, opts, callback) {
       'required': false
     });
   }
+  if (opts.includeWhiteboardUsages) {
```
nicolaasmatthijs: When an asset isn't included in an active whiteboard, it could still be part of an exported whiteboard, in which case the user should also not be able to delete it.

node_modules/col-assets/lib/api.js
nicolaasmatthijs: That makes sense. Unfortunately, showing the active whiteboards in which an asset is used isn't all that helpful as whiteboards are private and we wouldn't be able to send the user to the whiteboard. What would be useful is to include which exported whiteboards the asset is being used in, and I'd be happy with including that in the scope of this project. In terms of the logic associated to deciding whether a user can delete an asset, I'd prefer that to be consolidated in the back-end so we can update that logic without changing the UI.

node_modules/col-assets/lib/api.js
```diff
@@ -824,18 +835,26 @@ var deleteAsset = module.exports.deleteAsset = function(ctx, id, callback) {
...
+  var opts = {
+    'includeWhiteboardUsages': true
+  };
+  getAsset(ctx, id, opts, function(err, asset) {
     if (err) {
       return callback(err);
     }
 
+    // If an asset has no interactions, associated users may delete that asset. Other assets may be deleted
+    // only by course instructors
```
nicolaasmatthijs: I think we refer to them as `course administrators` elsewhere

node_modules/col-assets/tests/test-assets.js
```diff
@@ -881,9 +882,9 @@ describe('Assets', function() {
   describe('Delete asset', function() {
 
     /**
-     * Test that verifies that an asset can be deleted
+     * Test that verifies that an asset can be deleted by an instructor
```
nicolaasmatthijs: I would go with the `administrator` language here as well

node_modules/col-assets/tests/test-assets.js
```diff
@@ -920,6 +921,58 @@ describe('Assets', function() {
...
+                          // Verify that the creator cannot delete the liked asset
+                          AssetsTestUtil.assertDeleteAssetFails(client1, course, asset3.id, 401, function() {
+
+                            // Create an asset for use in a whiteboard
+                            AssetsTestUtil.assertCreateLink(client1, course, 'UC Riverside', 'http://www.ucr.edu/', null, function(asset4) {
+                              UsersTestUtil.assertGetMe(client1, course, null, function(user1Me) {
+                                WhiteboardsTestUtil.assertCreateWhiteboard(client1, course, 'UC Riverside Whiteboard', [user1Me.id], function(whiteboard) {
+                                  WhiteboardsTestUtil.addAssetToWhiteboard(client1, course, asset4, whiteboard, function() {
+                                    // Verify that the creator cannot delete the used asset
+                                    AssetsTestUtil.assertDeleteAssetFails(client1, course, asset4.id, 401, function() {
```
nicolaasmatthijs: Can we add a test case for an asset that's part of an exported whiteboard but no longer in use in an active whiteboard?

node_modules/col-whiteboards/tests/util.js
```diff
@@ -209,6 +209,36 @@ var addElementsToWhiteboard = module.exports.addElementsToWhiteboard = function(
...
+ * @param  {RestClient}         client                        The REST client to make the request with
+ * @param  {Course}             course                        The Canvas course in which the user is interacting with the API
+ * @param  {Whiteboard}         whiteboard                    The whiteboard to add the elements to
+ * @param  {Function}           callback                      Standard callback function
+ * @throws {AssertionError}                                   Error thrown when an assertion failed
+ */
+var addAssetToWhiteboard = module.exports.addAssetToWhiteboard = function(client, course, asset, whiteboard, callback) {
+  var assetElement = mockWhiteboardElement('path1.json', whiteboard.id);
+  assetElement.asset_id = asset.id;
+  DB.WhiteboardElement.upsert(assetElement).complete(function(err) {
```
nicolaasmatthijs: Can you also add a TODO here to replace this with the websocket approach ?

node_modules/col-whiteboards/tests/util.js
```diff
@@ -209,6 +209,36 @@ var addElementsToWhiteboard = module.exports.addElementsToWhiteboard = function(
   });
 };
 
+
+/**
+ * Add an asset to a whiteboard
+ *
+ * @param  {RestClient}         client                        The REST client to make the request with
+ * @param  {Course}             course                        The Canvas course in which the user is interacting with the API
```
nicolaasmatthijs: `asset` is not documented

node_modules/col-whiteboards/tests/util.js
```diff
@@ -209,6 +209,36 @@ var addElementsToWhiteboard = module.exports.addElementsToWhiteboard = function(
...
+ * @param  {Whiteboard}         whiteboard                    The whiteboard to add the elements to
```
nicolaasmatthijs: We're only adding a single element

node_modules/col-core/lib/db.js
```diff
@@ -638,6 +639,10 @@ var setUpModel = function(sequelize) {
   // A whiteboard can have multiple whiteboard elements
   Whiteboard.hasMany(WhiteboardElement, {'as': 'WhiteboardElements'});
 
+  // An asset can be used as an element in multiple whiteboards. Note that this relation differs from the
+  // AssetWhiteboardElement relation, which describes elements in whiteboards that have been exported as assets
+  Asset.hasMany(WhiteboardElement, {'as': 'whiteboard_usages'});
```
nicolaasmatthijs: Do we need a migration script for the production database?

node_modules/col-assets/tests/test-assets.js
```diff
@@ -920,6 +921,58 @@ describe('Assets', function() {
     });
 
     /**
+     * Test that verifies that an asset can be deleted by a user under certain conditions
```
nicolaasmatthijs: Can we replace this with `when the asset hasn't had any interactions yet` ?

node_modules/col-assets/lib/api.js
paulkerschen: Thanks @nicolaasmatthijs. The asset_whiteboard_elements table for exported whiteboards doesn't currently have a column for the asset id of the element (as opposed to asset id of the whiteboard). It's included in the JSON but not in a way that's queryable. Probably the quickest way to make an association possible is to add a column?

nicolaasmatthijs: Ah, that's unfortunate. Yeah, it's probably easiest to add a column. Note that we'll then also need to create a migration script for the existing data.

nicolaasmatthijs:

node_modules/col-core/lib/db.js
nicolaasmatthijs: That is probably correct. Defining these relationships always confuses me, so I just check the database to see if anything changed.

paulkerschen:
Follow-up pushed addressing PR comments and covering everything but the migration script. I'll need to check with @nicolaasmatthijs on convention for that, as I don't see previous migrations included in the project.

node_modules/col-assets/lib/api.js
```diff
@@ -907,6 +936,20 @@ var deleteAssets = module.exports.deleteAssets = function(ctx, assetIds, callbac
   });
 };
 
+var canDeleteAsset = function(user, asset) {
```
nicolaasmatthijs: Needs JSDoc above the function

node_modules/col-assets/lib/api.js
```diff
@@ -64,6 +67,13 @@ var getAssetProfile = module.exports.getAssetProfile = function(ctx, id, increme
 
     asset = asset.toJSON();
     asset.whiteboard_elements = _.pluck(asset.whiteboard_elements, ['element']);
+    asset.exported_whiteboards = _.pluck(asset.exported_whiteboards, ['id', 'title']);
```
nicolaasmatthijs: Do we need to do the `pluck` here as we're already doing `'attributes': ['title', 'id'],` below ?

node_modules/col-assets/lib/api.js
```diff
@@ -64,6 +67,13 @@ var getAssetProfile = module.exports.getAssetProfile = function(ctx, id, increme
...
+
+    // Usage in active whiteboards will prevent an asset from being deleted, but should not be kept in the asset profile
+    // once the canDelete property is set
```
nicolaasmatthijs: It's not entirely clear to me what this comment is trying to say

node_modules/col-assets/lib/api.js
```diff
@@ -824,18 +851,20 @@ var deleteAsset = module.exports.deleteAsset = function(ctx, id, callback) {
...
+  var opts = {
+    'includeWhiteboardUsages': true
+  };
+  getAsset(ctx, id, opts, function(err, asset) {
     if (err) {
       return callback(err);
     }
 
+    if (!canDeleteAsset(ctx.user, asset)) {
+      log.error({'id': id}, 'Unauthorized to delete an asset');
```
nicolaasmatthijs: Rather than just logging an `Unauthorized` message here, can you clarify that the asset can't be deleted because it has activity associated to it?

node_modules/col-assets/lib/api.js
```diff
@@ -907,6 +936,20 @@ var deleteAssets = module.exports.deleteAssets = function(ctx, assetIds, callbac
   });
 };
 
+var canDeleteAsset = function(user, asset) {
+  // If an asset has no interactions, associated users may delete that asset. Other assets may be deleted
+  // only by course administrators
+  if (user.is_admin) {
+    return true;
+  } else {
+    var isAssociatedUser = !!(_.find(asset.users, {'id': user.id}));
```
nicolaasmatthijs: Can we get rid of the double `!` ?

public/app/assetlibrary/item/item.html
```diff
@@ -4,17 +4,17 @@
 
   <!-- TITLE -->
   <div class="row assetlibrary-item-top">
-    <div class="col-xs-{{ me.is_admin ? 5 : (canManageAsset() ? 8 : 12) }} col-sm-{{ me.is_admin ? 7 : (canManageAsset() ? 10 : 12) }}">
+    <div class="col-xs-{{ asset.canDelete ? 5 : (canManageAsset() ? 8 : 12) }} col-sm-{{ asset.canDelete ? 7 : (canManageAsset() ? 10 : 12) }}">
```
nicolaasmatthijs: I think we've mostly been using the `can_delete` format in the REST API responses

nicolaasmatthijs: cfr. `is_admin` above

node_modules/col-whiteboards/tests/util.js
```diff
@@ -209,6 +209,57 @@ var addElementsToWhiteboard = module.exports.addElementsToWhiteboard = function(
...
+var removeAssetFromWhiteboard = module.exports.removeAssetFromWhiteboard = function(client, course, asset, whiteboard, callback) {
+  // TODO: Use the websocket rather than going through the database directly
+  var whiteboardElementOptions = {
+    'where': {
+      'asset_id': asset.id,
+      'whiteboard_id': whiteboard.id
+    }
+  };
+  DB.WhiteboardElement.destroy(whiteboardElementOptions).complete(function(err) {
+    assert.ok(!err);
```
nicolaasmatthijs: Can we also verify that the element doesn't exist anymore?

nicolaasmatthijs:
With regards to the database migration, we need to develop better migration script practices. However, for this PR, I would go with a folder inside the `scripts` folder using the format `yyyymmdd-colxxx`, containing:

- .sql file that contains required SQL query to alter database structure
- Node script that actively migrates existing data to the new schema

nicolaasmatthijs:

node_modules/col-assets/lib/api.js
paulkerschen: In fact we do need the `pluck`; because the association runs through whiteboard elements, the Sequelize call returns an array of whiteboard elements for each whiteboard, and specifying `attributes` below isn't sufficient to screen them out. We could remove the `attributes` option, but then the database will be returning many more unneeded asset-specific properties.

paulkerschen:
PR review addressed and migration script added; back to @nicolaasmatthijs.

scripts/20160525-col378/populate_element_asset_id.js
```diff
@@ -0,0 +1,112 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE. THE
+ * SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
+ * "AS IS". REGENTS HAS NO OBLIGATION TO PROVIDE MAINTENANCE, SUPPORT, UPDATES,
+ * ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+var _ = require('lodash');
+var async = require('async');
+var config = require('config');
+var CollabosphereConstants = require('col-core/lib/constants');
```
nicolaasmatthijs: This should be split into 2 blocks. I will make this change

scripts/20160525-col378/populate_element_asset_id.js
```diff
@@ -0,0 +1,112 @@
...
+ * Parse the serialized asset whiteboard element, populating the element_asset_id column if found
+ *
+ * @param  {AssetWhiteboardElement}       assetWhiteboardElement      The element to parse an asset id from
+ * @param  {Function}                     callback                    Standard callback function
+ * @param  {Object}                       callback.err                An error that occurred, if any
+ * @param  {Number}                       callback.assetId            Parsed asset id, if any 
+ */
+var populateElementAssetId = function(assetWhiteboardElement, callback) {
+  var assetId = assetWhiteboardElement.element.assetId;
+    if (assetId) {
```
nicolaasmatthijs: Indentation is off. I'll fix.

scripts/20160525-col378/populate_element_asset_id.js
```diff
@@ -0,0 +1,112 @@
...
+  });
+};
+
+/**
+ * Parse the serialized asset whiteboard element, populating the element_asset_id column if found
+ *
+ * @param  {AssetWhiteboardElement}       assetWhiteboardElement      The element to parse an asset id from
+ * @param  {Function}                     callback                    Standard callback function
+ * @param  {Object}                       callback.err                An error that occurred, if any
+ * @param  {Number}                       callback.assetId            Parsed asset id, if any
```
nicolaasmatthijs: Trailing whitespace. I'll fix.

nicolaasmatthijs:
Followed up and merged

PR #135, opened 2016-06-01
COL-387 Allow whiteboard search and filter - back end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-387 (back end component of https://jira.ets.berkeley.edu/jira/browse/COL-380)

Note the new analytics event 'Search whiteboards', which might require additional integration on the Mixpanel side.

### DIFF
--- a/node_modules/col-whiteboards/tests/test-whiteboards.js
+++ b/node_modules/col-whiteboards/tests/test-whiteboards.js
@@ -274,41 +274,41 @@ describe('Whiteboards', function() {
       setUpUsers(function(users, course, otherCourse) {
 
         // Retrieve the empty whiteboards list
-        WhiteboardsTestUtil.assertGetWhiteboards(users.creator.client, course, null, null, 0, function(whiteboards) {
+        WhiteboardsTestUtil.assertGetWhiteboards(users.creator.client, course, null, null, null, 0, function(whiteboards) {
 
           // Create a whiteboard
           WhiteboardsTestUtil.assertCreateWhiteboard(users.creator.client, course, 'UC Berkeley Whiteboard', users.regularA.me.id, function(whiteboard1) {
             // Verify that the whiteboard is returned for the creator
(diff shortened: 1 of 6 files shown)


node_modules/col-whiteboards/tests/test-whiteboards.js
```diff
@@ -450,6 +450,111 @@ describe('Whiteboards', function() {
...
+                    verifySearch(users.regularB.client, course, null, whiteboardsB.concat(whiteboardsShared), function() {
+
+                      // When keyword filters are applied, a regular user should see only their whiteboards with matching titles
+                      verifySearch(users.regularA.client, course, {'keywords': 'Berkeley'}, [privateBerkeleyBoard, sharedBerkeleyBoard], function() {
+                        verifySearch(users.regularB.client, course, {'keywords': 'Davis'}, [privateDavisBoard, sharedDavisBoard], function() {
+                          verifySearch(users.regularA.client, course, {'keywords': 'Davis'}, [sharedDavisBoard], function() {
+                            verifySearch(users.regularB.client, course, {'keywords': 'Berkeley'}, [sharedBerkeleyBoard], function() {
+
+                              // An administrator should be able to see all whiteboards associated with a user
+                              verifySearch(users.instructor.client, course, {'user': users.regularA.me.id}, whiteboardsA.concat(whiteboardsShared), function() {
```
nicolaasmatthijs: This implies that the search would return the 12 whiteboards that were created first and the new whiteboard, which means there would be 13 results. However, as we're not specifying a total number of results to return, we only expect 10.

This should be updated to verify that the search is returning the correct number of results as well.

nicolaasmatthijs: I noticed that the asset filter tests have the same issue. Can you implement a fix there as well?

nicolaasmatthijs:
This looks really good. Back to @paulkerschen for the comment I left.

paulkerschen:
Result paging verification added in asset and whiteboard searches; back to @nicolaasmatthijs.

nicolaasmatthijs:
Looks good. Merged.

PR #136, opened 2016-06-01
COL-352 Fix nested form in asset reuse modal

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-352

Note that the changes are mostly indentation.

### DIFF
--- a/public/app/assetlibrary/search/search.html
+++ b/public/app/assetlibrary/search/search.html
@@ -19,8 +19,8 @@
           </button>
         </span>
       </div>
-    </div>
-  </form>
+    </form>
+  </div>
 
   <!-- ADVANCED SEARCH -->
   <div data-ng-show="isAdvancedSearch" class="col-pane">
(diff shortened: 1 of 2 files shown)


nicolaasmatthijs:
@paulkerschen : It looks like this is tripping up the validation:

screen shot 2016-06-01 at 5 14 21 pm

nicolaasmatthijs:
Back to @paulkerschen for follow-up

paulkerschen:
@nicolaasmatthijs, looks like the problem is specific to Firefox and can be resolved by removing the `data-ng-required`on the inputs (as well as the `$valid` check on the submit button, which presumably is no longer doing anything). 

The `ng-disabled` on the submit button is still in place and prevents submission when no assets are selected. Does that seem acceptable?

paulkerschen:
Candidate fix pushed for the validation issue, back to @nicolaasmatthijs for review.

nicolaasmatthijs:
Looks good. Merged.

PR #141, opened 2016-07-18
COL-403 Include asset_id when saving comment activities

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-403

### DIFF
--- a/node_modules/col-activities/lib/api.js
+++ b/node_modules/col-activities/lib/api.js
@@ -177,13 +177,13 @@ var exportActivities = module.exports.exportActivities = function(ctx, callback)
  * @param  {Number}           objectId                            The id of the object on which the activity is taking place (e.g., the asset id, the comment id, etc.)
  * @param  {String}           objectType                          The type of the object on which the activity is taking place. One of `CollabosphereConstants.ACTIVITY.OBJECT_TYPES`
  * @param  {Object}           [metadata]                          Additional metadata that is associated with the activity. For example, when creating a submission activity it might be useful to store the submission id, the attempt number and the attachment ids
+ * @param  {Number}           [metadata.assetId]                  The id of an asset associated with the activity, when different from the object id (e.g., for comments)
  * @param  {User}             [actor]                             The user performing the activity when different than the user earning activity points
- * @param  {Number}           [assetId]                           The associated asset id when different from the object id (e.g., for comments)
  * @param  {Function}         callback                            Standard callback function
  * @param  {Object}           callback.err                        An error that occurred, if any
  * @param  {Activity}         callback.activity                   The created activity
  */
-var createActivity = module.exports.createActivity = function(course, user, type, objectId, objectType, metadata, actor, assetId, callback) {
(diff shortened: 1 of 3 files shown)


paulkerschen:
Assigned to @nicolaasmatthijs for review.

node_modules/col-activities/lib/api.js
```diff
@@ -178,26 +178,29 @@ var exportActivities = module.exports.exportActivities = function(ctx, callback)
...
  * @param  {Object}           [metadata]                          Additional metadata that is associated with the activity. For example, when creating a submission activity it might be useful to store the submission id, the attempt number and the attachment ids
  * @param  {User}             [actor]                             The user performing the activity when different than the user earning activity points
+ * @param  {Number}           [assetId]                           The associated asset id when different from the object id (e.g., for comments)
  * @param  {Function}         callback                            Standard callback function
  * @param  {Object}           callback.err                        An error that occurred, if any
  * @param  {Activity}         callback.activity                   The created activity
  */
-var createActivity = module.exports.createActivity = function(course, user, type, objectId, objectType, metadata, actor, callback) {
+var createActivity = module.exports.createActivity = function(course, user, type, objectId, objectType, metadata, actor, assetId, callback) {
```
nicolaasmatthijs: This is intended to be a fairly generic function that deals with generating activities. There's certain activity types (e.g., whiteboard activities) for which `assetId` doesn't really make sense as part of the signature. Therefore, I'd suggest storing the `asset_id` as part  of the activity metadata.

nicolaasmatthijs:

node_modules/col-activities/lib/api.js
paulkerschen: @nicolaasmatthijs If we don't want to change the function signature, I'm fine with passing the assetId in as part of the metadata; but in that case we'd still want to extract it within the function and store it in its own column to enable SQL joins. Make sense?

nicolaasmatthijs: I think it's technically possible in Postgres to join by a value inside of a stored JSON document, but I'm not entirely sure how performant this is. The alternative could be to just generalize `assetId` a bit more (e.g., `targetId`) so it can apply to whiteboards as well?

paulkerschen: @nicolaasmatthijs We could; my concern is that the activities table already has an `asset_id` column separate from the `object_id` column, which just happens to go unused for this set of comment activities. I agree it would be more elegant down the road to either try Postgres JSON join or generalize `assetId` to `targetId`, but since this would involve refactoring existing code that already depends on the `assetId` join, I think that would be better left to a follow-up task.

nicolaasmatthijs: Sorry, I missed the fact that we already have an `asset_id` column. In that case, let's proceed with adding it to the metadata and then pulling it out to store explicitly.

paulkerschen: Will do, @nicolaasmatthijs.

paulkerschen:
Review addressed, back to @nicolaasmatthijs.

nicolaasmatthijs:
Looks good. Merged.

PR #147, opened 2016-07-28
COL-415, remove selenium/ dir

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-415

cc: @pfarestveit


nicolaasmatthijs:
Looks good. Merged.

PR #151, opened 2016-08-10
COL-406 COL-407 Add new whiteboard-related activity types

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-406
https://jira.ets.berkeley.edu/jira/browse/COL-407

Like other activity types, these are viewable and modifiable through the Engagement Index.

Tested locally, but no integration tests added as our suite currently doesn't cover whiteboard actions through websockets.

### DIFF
--- a/node_modules/col-whiteboards/lib/api.js
+++ b/node_modules/col-whiteboards/lib/api.js
@@ -39,6 +39,7 @@ var Sequelize = require('sequelize');
 var spawn = require('child_process').spawn;
 var util = require('util');
 
+var ActivitiesAPI = require('col-activities');
 var AnalyticsAPI = require('col-analytics');
 var AssetsAPI = require('col-assets');
 var CanvasAPI = require('col-canvas');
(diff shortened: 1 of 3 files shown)


node_modules/col-whiteboards/lib/api.js
```diff
@@ -755,7 +756,24 @@ var whiteboardAddActivity = function(socket, elements, callback) {
       AnalyticsAPI.track(socket.ctx.user, 'Add whiteboard element', AnalyticsAPI.getWhiteboardElementProperties(element, socket.whiteboard, elements.length));
     });
 
-    return callback();
+    async.each(elements, function(element, done) {
```
nicolaasmatthijs: I'm a little bit concerned about the `ActivitiesAPI.createActivity` function being executed in parallel, as students could be adding many assets at once. I believe that function adds the activity and recalculates and caches the total score for that user, and I wonder if that could suffer from a race condition.

nicolaasmatthijs:

node_modules/col-whiteboards/lib/api.js
paulkerschen: @nicolaasmatthijs Thanks, would `eachSeries` be the proper way to do it serially?

nicolaasmatthijs: That would be the right way to do them serially, so might be the safest thing to do here.

A potentially better would be to allow `ActivitiesAPI.createActivity` to process multiple activities at once, but that might be out of scope for this PR.

nicolaasmatthijs:
Looks good. Merged

PR #152, opened 2016-08-11
COL-400 Calculate current user ranks for weekly email

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-400

Since we need both points and email addresses for users, the function signature of UsersAPI.getLeaderboard is expanded to include an `options` object, following the pattern of UsersAPI.getAllUsers. These changes are opaque to our current integration test suite, but will need to be covered in integration tests for the weekly email.

The various permutations for rank calculation will also need test coverage.

### DIFF
--- a/node_modules/col-activities/lib/notifications/weekly.js
+++ b/node_modules/col-activities/lib/notifications/weekly.js
@@ -119,8 +119,6 @@ var handleUser = function(course, activitySummary, user, callback) {
   }
 
   // Parse out activity data specific to this user.
-  // TODO: Include ranks for the user: last week's and this week's both.
-
   var userData = activitySummary.users[user.id] || {};
   if (!_.isEmpty(userData)) {
     // Find the user's most popular asset for the last week. We define asset 'popularity' as a weighted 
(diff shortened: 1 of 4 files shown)


node_modules/col-users/lib/api.js
```diff
@@ -312,11 +312,25 @@ var getAllUsers = module.exports.getAllUsers = function(ctx, options, callback)
...
+var getLeaderboard = module.exports.getLeaderboard = function(ctx, options, callback) {
+  options = options || {};
+  options.enrollmentStates = options.enrollmentStates || [
+    CollabosphereConstants.ENROLLMENT_STATE.ACTIVE,
+    CollabosphereConstants.ENROLLMENT_STATE.INVITED
+  ];
+
+  var attributes = UserConstants.POINTS_USER_FIELDS;
+  if (options.includeEmail) {
+    attributes = _.union(UserConstants.POINTS_USER_FIELDS, ['canvas_email']);
```
nicolaasmatthijs: Maybe `canvas_email` should be pulled out into a constant as well

node_modules/col-users/lib/api.js
```diff
@@ -312,11 +312,25 @@ var getAllUsers = module.exports.getAllUsers = function(ctx, options, callback)
  * Get the users in the current course and their points
  *
  * @param  {Context}    ctx                             Standard context containing the current user and the current course
+ * @param  {Object}     [options]                       Filter options
+ * @param  {String[]}   [options.enrollmentStates]      The enrollment states of the users in the course. One of {@link CollabosphereConstants.ENROLLMENT_STATE}. Defaults to `active` and `invited`
+ * @param  {String}     [options.includeEmail]          Whether to include the email address on the user objects. It's up to the caller to ensure email addresses are dealt with securely
```
nicolaasmatthijs: Can we indicate that the email address will not be included by default?

node_modules/col-activities/lib/notifications/weekly.js
```diff
@@ -184,24 +182,37 @@ var getCourseData = function(course, callback) {
...
+      }
+      return _.extend(user.toJSON(), {
+        'rank': {
+          'thisWeek': rank
+        }
+      });
+    });
+
+    // Index users by id for quick lookup.  
+    users = _.indexBy(users, 'id');
```
nicolaasmatthijs: We say `so rank can be derived from array index` above, but then immediately revert it back to an object

node_modules/col-activities/lib/notifications/weekly.js
```diff
@@ -184,24 +182,37 @@ var getCourseData = function(course, callback) {
...
-    users = _.chain(users)
-      .map(function(user) {
-        return user.toJSON();
-      })
-      .indexBy('id')
-      .value();
+    // Calculate user ranks. Users are returned in descending order of score, so rank can be derived from array index. Rank 
+    // is defined as the number of users with a higher score, plus one; e.g., scores [200, 100, 50, 50, 25] map to ranks
+    // [1, 2, 3, 3, 5].
+    // TODO: Store this week's ranks and retrieve last week's ranks.
```
nicolaasmatthijs: If we will be storing this week's and last week's rank, is it necessary to calculate the rank here?

nicolaasmatthijs:

node_modules/col-activities/lib/notifications/weekly.js
paulkerschen: @nicolaasmatthijs It makes sense to store the rank here immediately after calculation, so that "this week's rank," as sent to the user, is guaranteed to be the same when we retrieve it a week from now as "last week's rank." The alternative would be some separate weekly process to calculate and store ranks, but that doesn't seem necessary.

paulkerschen:
Comments addressed, back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #154, opened 2016-08-17
COL-433 Fix error logging in whiteboards

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-433


johncrossman:
Thanks @paulkerschen  Looks good.

PR #155, opened 2016-08-18
COL-383, set process.exitCode instead of calling process.exit() directly

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-383

See https://nodejs.org/api/process.html regarding misuse of `process.exit()`.  I want to be sure we're not losing info that would be otherwise be written to stdout. See best practice described in URL above. Excerpt:
```
It is important to note that calling process.exit() will force the process to exit 
as quickly as possible even if there are still asynchronous operations pending 
that have not yet completed fully, including I/O operations to process.stdout 
and process.stderr.
```

### DIFF
--- a/node_modules/col-whiteboards/data/whiteboardToPng.js
+++ b/node_modules/col-whiteboards/data/whiteboardToPng.js
@@ -27,6 +27,7 @@
 
 var _ = require('lodash');
 var fabric = require('fabric').fabric;
+var log = require('col-core/lib/logger')('col-whiteboards/data/whiteboardToPng');
 
 // The amount of padding (in pixels) that should be added to whiteboards that get exported to PNG
 var WHITEBOARD_PADDING = 10;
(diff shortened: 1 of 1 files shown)


paulkerschen:
@johncrossman I'm not sure about this, but if we're not calling exit() explicitly, we might need to pass the error farther upstream via callbacks after setting the exit code and logging. See the end of the linked doc: `If it is necessary to terminate the Node.js process due to an error condition, throwing an uncaught error and allowing the process to terminate accordingly is safer than calling process.exit()` The question is whether this whiteboard error is in fact serious enough to warrant terminating the process.

johncrossman:
@paulkerschen Good point. And thanks for the excerpt. I'll reproduce locally.

paulkerschen:
@johncrossman Oh, I didn't realize this was all happening inside a spawned child process. Perhaps fine.

node_modules/col-whiteboards/data/whiteboardToPng.js
```diff
@@ -42,15 +42,17 @@ var init = function() {
 
     // Create a PNG stream for these elements
     generatePng(whiteboardElements, function(err, data, dimensions) {
+      // According to https://nodejs.org/api/process.html, "rather than calling process.exit() directly,
```
nicolaasmatthijs: The convention is to use `@see` for external references, so this would become something along the following lines:

```
// Rather than calling process.exit() directly, set the process.exitCode and allow the process to exit naturally
// @see https://nodejs.org/api/process.html
```

node_modules/col-whiteboards/data/whiteboardToPng.js
```diff
@@ -42,15 +42,17 @@ var init = function() {
...
+      // the code should set the process.exitCode and allow the process to exit naturally"
       if (err) {
         console.error(err);
-        process.exit(2);
+        process.exitCode = 2;
```
nicolaasmatthijs: I don't think setting the `exitCode` is going to fix anything in this case. The process is being explicitly exited because there's nothing else left to do.

node_modules/col-whiteboards/data/whiteboardToPng.js
```diff
@@ -42,15 +42,17 @@ var init = function() {
 
     // Create a PNG stream for these elements
     generatePng(whiteboardElements, function(err, data, dimensions) {
+      // According to https://nodejs.org/api/process.html, "rather than calling process.exit() directly,
+      // the code should set the process.exitCode and allow the process to exit naturally"
       if (err) {
         console.error(err);
```
nicolaasmatthijs: One thing that might be useful to do here is to hook this up to the standard SuiteC logging to ensure that this log message can be found in the application log.

nicolaasmatthijs:
@johncrossman : Were you able to reproduce the bug locally and did this fix resolve the problem? My suspicion was that this bug is caused by Fabric when rendering a whiteboard in the back-end. If that's true, it would make sense to hold off on this ticket until the Node upgrade has been completed and merged, as that might solve the problem.

paulkerschen:
@johncrossman I think hooking up to the standard Collabosphere logger is a good idea: adding to the top of the file something like `var log = require('col-core/lib/logger')('col-whiteboards/data/whiteboardToPng');` should enable you to send nicely formatted messages to the app log. With that (and if you like the suggestion on `@see` in JSDocs) we can then try this out.

johncrossman:
Following up on this PR now.

johncrossman:
Closing w/ the hope that recent Node upgrade will fix the problem. If that is not true, we'll consider options.

PR #159, opened 2016-08-22
COL-449, migrate_assets.js script creates dedicated temp dir for file downloads

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-449

Migrate file knowing that temp dir creation succeeded. Significant change is:
```
var downloadDir = path.join(os.tmpdir(), processStartTime.toString());
fs.stat(downloadDir, function(err, stat) {
  ...
```
Add `?w=1` to URL to ignore whitespace in diff.

### DIFF
--- a/scripts/migrate_assets.js
+++ b/scripts/migrate_assets.js
@@ -30,6 +30,8 @@ var async = require('async');
 var contentDisposition = require('content-disposition');
 var fs = require('fs');
 var os = require('os');
+var fs = require('fs');
+var path = require('path');
 var request = require('request');
 var argv = require('yargs')
     .usage('Usage: $0 --fromcourse [fromcourse] --fromuser [fromuser] --tocourse [tocourse] --touser [touser] --categories [categories]')
@@ -57,6 +59,7 @@ var fromUser = argv.fromuser;
 var toCourse = argv.tocourse;
 var toUser = argv.touser;
 
+var processStartTime = Date.now();
 // Variable that will keep track of the source user context for interaction with the API
 var fromCtx = null;
 // Variable that will keep track of the destination user context for interaction with the API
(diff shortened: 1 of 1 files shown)


scripts/migrate_assets.js
```diff
@@ -318,39 +321,50 @@ var migrateLink = function(link, callback) {
...
-    };
+  var downloadDir = path.join(os.tmpdir(), processStartTime.toString());
+  fs.stat(downloadDir, function(err, stat) {
+    // Download the file to a temporary folder
+    if (err !== null) {
+      if (err.code === 'ENOENT') {
+        fs.mkdirSync(downloadDir);
+      } else {
+        log.error({downloadDir: downloadDir, errorCode: err.code}, 'Failed to create temp directory during file migration');
+        callback(err);
```
paulkerschen: @johncrossman Does this need to be `return callback(err)` so that code farther down doesn't execute?

scripts/migrate_assets.js
```diff
@@ -318,39 +321,50 @@ var migrateLink = function(link, callback) {
...
-    // Function that will clean up the temporary file
-    var cleanTempFile = function(err) {
-      fs.unlink(path, function() {
-        return callback(err);
-      });
-    };
+  var downloadDir = path.join(os.tmpdir(), processStartTime.toString());
+  fs.stat(downloadDir, function(err, stat) {
+    // Download the file to a temporary folder
+    if (err !== null) {
```
paulkerschen: @johncrossman It looks like `if (err) {` is the usual idiom in the codebase, probably for purposes of catching both `null` and `undefined`.

johncrossman: Thanks @paulkerschen  Amending.

scripts/migrate_assets.js
johncrossman: Yes, indeed. Good catch.

paulkerschen:
Thanks @johncrossman.

PR #160, opened 2016-08-22
COL-453 Don't attempt to double-delete temp files when processing Canvas assignments

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-453


johncrossman:
Thanks, @paulkerschen  Looks clean.

PR #161, opened 2016-08-22
COL-307, /api/status now reports app version info

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-307

`/api/status` will have:
```
{
app: true,
db: true,
version: "1.6",
...
}
```

### DIFF
--- a/node_modules/col-core/lib/api.js
+++ b/node_modules/col-core/lib/api.js
@@ -30,6 +30,7 @@ var express = require('express');
 var fs = require('fs');
 var os = require('os');
 var path = require('path');
+var appPackage = require('../../../package.json');
 var request = require('request');
 var util = require('util');
 
@@ -348,7 +349,8 @@ var initializeCanvasPoller = function() {
 var getStatus = module.exports.getStatus = function(callback) {
   // Indicate that the Node.js process is up
   var status = {
-    'app': true
+    'app': true,
+    'version': appPackage.version
   };
 
   // Check if a connection with the database can be made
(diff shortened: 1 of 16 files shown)


versions/application.txt
```diff
@@ -0,0 +1 @@
+1.6
```
nicolaasmatthijs: Rather than having a separate file with the version number, a better approach might be to have an accurate version number in the `package.json` file and read the version number from that.

node_modules/col-core/lib/api.js
```diff
@@ -355,6 +355,16 @@ var getStatus = module.exports.getStatus = function(callback) {
   CanvasAPI.getCanvases(function(err, canvases) {
     status.db = err ? false : true;
 
+    // Application version number
+    var versionsFile = path.join(__dirname, '../../..', 'versions/application.txt');
+    fs.readFile(versionsFile, "utf8", function(err, data) {
```
paulkerschen: @johncrossman Since this is asynchronous, I think the function might exit before the file is read. You'd either want to use `readFileSync` or put the rest of the `getStatus` code inside the callback.

paulkerschen:
@johncrossman Perhaps worth adding a line to `col-core/tests/test-status.js` asserting a non-empty version property?

johncrossman:
@paulkerschen I've amended this PR such that version is in package.json.  I'll now add to `col-core/tests/test-status.js`.

johncrossman:
Amendments are in.  cc: @paulkerschen

paulkerschen:
Good news for Ops. Thanks @johncrossman.

PR #162, opened 2016-08-23
COL-425 Allow course admins to manually trigger weekly email from the browser

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-425

An admin-only magic URL is exposed to trigger the weekly email. This URL is also used for the first integration tests; more to come. 

Note that many utility functions are moved from `node_modules/col-activities/tests/test-daily.js` into `node_modules/col-activities/tests/util.js` so that both daily and the weekly email tests can use them.

### DIFF
--- a/config/travis.js
+++ b/config/travis.js
@@ -46,6 +46,8 @@ module.exports = {
   },
   "email": {
     "enabled": false,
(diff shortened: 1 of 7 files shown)


johncrossman:
Looks good. Thanks, @paulkerschen

PR #163, opened 2016-08-24
COL-434 Add per-course engagementindex_url property

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-434

The weekly email requires a stored URL for the engagement index, as we already do for the other two SuiteC tools.


johncrossman:
Thank you, @paulkerschen  Looks mighty fine.

PR #167, opened 2016-08-31
COL-472 Move version number reporting to its own feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-472

For Nagios reasons, Ops prefers that the /api/status feed not be subject to regular change.


johncrossman:
Are job is to keep Ops happy. Thanks, @paulkerschen

PR #168, opened 2016-08-31
COL-397 Add radar chart to weekly email

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-397

We generate one chart per user per week, and for the moment are simply storing them as PNG files on disk in a directory accessible to Apache. Generated files are between 17-20 kB; a script to delete older files after a certain point might be an advisable follow-up.


johncrossman:
I see no problems here, @paulkerschen  Nice work.

PR #170, opened 2016-09-01
COL-457 Add a poller health check to /api/status

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-457


johncrossman:
A healthy PR. Thx, @paulkerschen

PR #171, opened 2016-09-02
COL-444 Schedule weekly email on app startup

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-444

Per user request, weekly emails are set to Monday morning by default.

Includes a second commit to skip the weekly email for courses where the asset library or engagement index is not enabled.

### DIFF
--- a/node_modules/col-activities/tests/util.js
+++ b/node_modules/col-activities/tests/util.js
@@ -892,6 +892,40 @@ var assertEmailsForUsers = function(emails, users) {
 };
 
 /**
+ * Assert that a request to send weekly notification emails passes authorization but sends no messages
+ *
+ * @param  {RestClient}          client                           The REST client to make the request with
+ * @param  {Course}              course                           The Canvas course in which the user is interacting with the API
+ * @param  {Course}              dbCourse                         The database model for the course
+ * @param  {User[]}              users                            The users who should receive an email
+ * @param  {Function}            callback                         Standard callback function
+ * @param  {Object[]}            callback.emails                  The sent email messages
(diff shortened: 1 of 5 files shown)


node_modules/col-activities/tests/util.js
```diff
@@ -892,6 +892,40 @@ var assertEmailsForUsers = function(emails, users) {
...
+ *
+ * @param  {RestClient}          client                           The REST client to make the request with
+ * @param  {Course}              course                           The Canvas course in which the user is interacting with the API
+ * @param  {Course}              dbCourse                         The database model for the course
+ * @param  {User[]}              users                            The users who should receive an email
+ * @param  {Function}            callback                         Standard callback function
+ * @param  {Object[]}            callback.emails                  The sent email messages
+ * @throws {AssertionError}                                       Error thrown when an assertion fails
+ */
+var assertSendWeeklyNotificationsEmpty = module.exports.assertSendWeeklyNotificationsEmpty = function(client, course, dbCourse, users, callback) {
```
johncrossman: @paulkerschen I'm curious why this assertion doesn't live in `test-weekly.js`.

paulkerschen: @johncrossman That's a good question. The convention in the codebase seems to be that these multipart assertions mostly live in `util.js` files, even if they're only used in one place. This allows the `test-*.js` files to function as slimmer descriptions of application flow (which is often fairly complex, since they're all integration tests).

johncrossman:
Looks good, @paulkerschen  Thanks for the instructive test coverage.

PR #176, opened 2016-09-07
COL-476 Weekly email: use darker light gray

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-476

These email templates have no access to any sass-type extension that would let us define a constant.

### DIFF
--- a/email-templates/_shared/style.css
+++ b/email-templates/_shared/style.css
@@ -3,7 +3,7 @@
  *************/
 
 body {
-  background-color: #F2F2F2;
+  background-color: #d6d6d6;
   color: #333;
   font-size: 14px;
   height: 100%;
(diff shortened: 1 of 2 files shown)


email-templates/_shared/style.css
```diff
@@ -3,7 +3,7 @@
  *************/
 
 body {
-  background-color: #F2F2F2;
+  background-color: #d6d6d6;
```
johncrossman: @paulkerschen Thanks for the lowercase hex colors. IMO, easier to read.

johncrossman:
Looks good. Waiting on T

email-templates/_shared/style.css
nicolaasmatthijs: Note that the convention in the UI has been uppercase hex colors, and I believe CSS lint also verifies that (or at least did at some point)

PR #149, opened 2016-08-04
Upgrade to latest version of Node

nicolaasmatthijs:
https://jira.ets.berkeley.edu/jira/browse/COL-417

PR #184, opened 2016-09-12
COL-417 Upgrade to Node 6

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-417

Finalized version of #149. Apart from config file changes, we switch out deprecated methods and include a couple of a minor code tweaks to fix intermittent test failures.


node_modules/col-assets/lib/api.js
```diff
@@ -1265,7 +1265,7 @@ var createComment = module.exports.createComment = function(ctx, assetId, body,
           }
 
           // Retrieve the created comment, including the associated user
-          return getComment(comment.id, function(err, comment) {
+          getComment(comment.id, function(err, comment) {
```
johncrossman: @paulkerschen `return` is no longer needed?

.travis.yml
```diff
@@ -1,13 +1,18 @@
 sudo: required
 language: node_js
 node_js:
-  - "0.10.25"
+  - "6"
```
johncrossman: @paulkerschen this might be a config trick I don't know. Does "6" give us a snapshot of the latest stable 6.* release?  I would expect "6.5.0" here.

paulkerschen: @johncrossman - exactly, latest stable 6.* https://docs.travis-ci.com/user/languages/javascript-with-nodejs

We could nail it down more specifically, but since this is a test-only environment it might be good to have advance warning of breaking changes.

johncrossman:
Looks good, @paulkerschen  I put a couple questions. Ready to merge upon your input.

node_modules/col-assets/lib/api.js
paulkerschen: @johncrossman - Yes, this `return` was superfluous as the real `return` happens one level deeper inside the function.

PR #186, opened 2016-09-12
COL-491 Fix syntax updates in whiteboard API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-491

col-assets and col-whiteboards need better Nouveau Lodash.

Also - while I'm unaware of this causing actual bugs  - Lodash syntax updates in front-end code are reverted for the moment. The front end is using a separate, earlier version of Lodash (as defined in bower.json) and updating that should be a distinct task.


johncrossman:
+1  And thank u for the explanation, @paulkerschen

PR #187, opened 2016-09-14
COL-383 Better buffering for whiteboard PNG export

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-383


johncrossman:
I approve of any code with the comment `...waiting on the 'drain' event`.  I think Parker had that in Ireland.

PR #188, opened 2016-09-15
COL-417 Roll back to Node v0.10

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-417

With regret, the core of #184 is rolled back for infrastructure reasons. Most of our Node dependencies are rolled back too; however, we keep Lodash 4 because it's compatible, and because it was the portion of the upgrade responsible for code changes.


johncrossman:
image

PR #189, opened 2016-09-20
COL-398 - Make assignment synchronization configurable by assignment

nicolaasmatthijs:
This PR removes the per course toggle for synchronizing assignment submissions and adds a per assignment toggle that determines whether assignment submissions for that assignments will synchronize with the Asset Library. This can be changed by the instructor at any time.

Note 1: By default, assignment categories will not synchronize with the asset library. We'll need to think about how we want to do the migration. We could set all existing assignment categories to visible, as that would replicate the current functionality (except for those that have used the now removed toggle), or we could set them all to invisible, but then instructors would need to go in and enable the categories. We could do a more advanced migration script that looks at the setting that's being removed to determine what the default for that course should be.

Note 2: While the tests in this branch are succeeding, there's still some new tests that are missing. I'll try to update this PR later today.


nicolaasmatthijs:
Note 3: Added tests for listing categories and the default assignment category visibility. I want to add one more tests that verifies listing assets itself.

nicolaasmatthijs:
All of the tests that I wanted to add are now in place. This PR should be ready for review.

node_modules/col-assets/lib/api.js
```diff
@@ -281,133 +281,126 @@ var getAssets = module.exports.getAssets = function(ctx, filters, limit, offset,
...
+                 'LEFT JOIN asset_users au ON a.id = au.asset_id ' +
+                 'LEFT JOIN activities act ON a.id = act.asset_id' +
+                 ' AND act.course_id = :course_id' +
+                 ' AND act.user_id = :user_id' +
+                 ' AND act.object_type = \'asset\'' +
+                 ' AND act.type IN(\'like\', \'dislike\') ' +
+                 'WHERE a.deleted_at IS NULL' +
+                 ' AND a.course_id = :course_id ' +
+                 ' AND a.visible = TRUE ' +
+                 ' AND (c.visible = TRUE OR c.visible IS NULL)';
```
johncrossman: @nicolaasmatthijs your alter to `categories` table has `NOT NULL` constraint on the new `visible` column. Can we ditch the slightly confusing `...OR c.visible IS NULL)` above? Are you trying to account for legacy data?

node_modules/col-assets/lib/api.js
```diff
@@ -281,133 +281,126 @@ var getAssets = module.exports.getAssets = function(ctx, filters, limit, offset,
...
-          '$iLike': '%' + keywords + '%'
-        },
-        'description': {
-          '$iLike': '%' + keywords + '%'
-        }
-      }
-    });
+    var keywords = '%' + filters.keywords.trim().replace(/ /g, '%') + '%';
+    sqlQuery += ' AND (a.title ILIKE :keywords OR a.description ILIKE :keywords)';
+    params['keywords'] = keywords;
```
johncrossman: @nicolaasmatthijs a little streamlining is possible:
`params['keywords'] =  '%' + filters.keywords.trim().replace(/ /g, '%') + '%';`
No need to declare `var keywords`, right?

node_modules/col-canvas/lib/poller.js
```diff
@@ -464,9 +469,11 @@ var handleAssignmentCategory = function(ctx, users, categories, assignment, call
       return callback();
     });
 
-  // Create a category for an assignment if it doesn't exist yet
+  // Create a category for an assignment if it doesn't exist yet.
+  // By default, the category will be marked as not visible
   } else {
-    CategoriesAPI.createCategory(ctx, assignment.name, assignment.id, assignment.name, function(err, newCategory) {
+    var enableAssignmentCategories = opts.enableAssignmentCategories  || false;
```
johncrossman: @nicolaasmatthijs extra space before `||`

node_modules/col-assets/lib/api.js
nicolaasmatthijs: While every category is required to have a `visible` column, not every asset will be associated to a category. In order to ensure that assets not associated to a category are also returned in the list, the `c.visible IS NULL` is required.

node_modules/col-assets/lib/api.js
nicolaasmatthijs: Yep, good point. Should be fixed now.

node_modules/col-canvas/lib/poller.js
nicolaasmatthijs: Good catch. Fixed.

node_modules/col-categories/lib/api.js
```diff
@@ -148,22 +158,28 @@ var getCategoriesById = module.exports.getCategoriesById = function(ctx, ids, ca
...
  * @param  {Number}         [canvasAssignmentId]        The id of the assignment to which this category is linked
  * @param  {String}         [canvasAssignmentName]      The name of the assignment in Canvas to which this category is linked
  * @param  {Function}       callback                    Standard callback function
  * @param  {Object}         callback.err                An error that occurred, if any
  * @param  {Category}       callback.category           The created category
  */
-var createCategory = module.exports.createCategory = function(ctx, title, canvasAssignmentId, canvasAssignmentName, callback) {
+var createCategory = module.exports.createCategory = function(ctx, title, visible, canvasAssignmentId, canvasAssignmentName, callback) {
+  // Default the visible setting
+  visible = (visible === false) ? false : true;
```
johncrossman: @nicolaasmatthijs how about `visible = !!visible`?  Bang, bang!

nicolaasmatthijs: I personally find `(visible === false) ? false : true;` to be a bit more explicit about what's going to happen. I don't feel strongly about that though, so happy to change it if needed.

node_modules/col-canvas/lib/poller.js
johncrossman: @nicolaasmatthijs for what it's worth, the line above seems unnecessary. Your logic in `CategoriesAPI.createCategory` would convert falsey to false.

node_modules/col-categories/lib/api.js
johncrossman: Well, let's leave it as is. The pattern is used elsewhere and consistency is more important. Thx

johncrossman:
@nicolaasmatthijs @paulkerschen Per John Scott and Oliver, this feature is scheduled for Spring 2017. Do not merge until:

1. we have the exact release schedule worked out and 
1. the migration concern in 'Note 1' (above) is answered

node_modules/col-canvas/lib/poller.js
nicolaasmatthijs: By default, a category will be created with a visibility of `true`. However, when a category is created from a Canvas assignment by the poller, its visibility is set to `false` by default. Therefore, the line above is necessary to ensure that `enableAssignmentCategories` isn't set to `null`, as the category would then be created as `visible`

paulkerschen:
@nicolaasmatthijs Sorry that we've let review on this languish, and in the meantime conflicts have crept in. Would you have time to take a pass at resolving the conflicts? If so, we'll get this merged in ASAP.

nicolaasmatthijs:
@paulkerschen : Ah, yes, I had forgotten about this in the midst of all the travel. I'll look into this first.

nicolaasmatthijs:
Closing this PR as I'm no longer allowed to push into the ETS repos

nicolaasmatthijs:
@johncrossman @paulkerschen : The updated PR is available at https://github.com/nicolaasmatthijs/collabosphere/pull/6. However, I'm currently not able to submit a PR against the ETS repo, so have sent it against my own. If you want to update my permissions, I can send the PR against the ETS repo.

johncrossman:
@nicolaasmatthijs Try again. I've updated your Github perms.

nicolaasmatthijs:
Thx! The updated PR is now available at https://github.com/ets-berkeley-edu/collabosphere/pull/201

PR #201, opened 2016-11-10
COL-398 - Make assignment synchronization configurable by assignment

nicolaasmatthijs:
This PR removes the per course toggle for synchronizing assignment submissions and adds a per assignment toggle that determines whether assignment submissions for that assignments will synchronize with the Asset Library. This can be changed by the instructor at any time.

Note: By default, assignment categories will not synchronize with the asset library.

1. We'll need to think about how we want to do the migration. We could set all existing assignment categories to visible, as that would replicate the current functionality (except for those that have used the now removed toggle), or we could set them all to invisible, but then instructors would need to go in and enable the categories. We could do a more advanced migration script that looks at the setting that's being removed to determine what the default for that course should be.

2. Defaulting to not synchronizating for every new assignment is the safest option, but requires the instructor to go in and enable it. As assignments aren't created too frequently, maybe that's acceptable. Might want to check with John Scott whether it is or whether we need a configurable default for every course as well.


paulkerschen:
Thank you for this, @nicolaasmatthijs, and for the notes on migration and configuration. We'll confer with our customer on how he wants to approach it.

PR #202, opened 2016-11-10
COL-473 Award EI points for multiple postings to a topic

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-473

Once this new logic is in place, the poller should gamely sweep through and pick up older discussion activities in the new format. Discussion activities in the old format will appear as duplicates. The migration script to remove those duplicates should be run post-release, once we've confirmed that the new poller logic is working as expected.


johncrossman:
@paulkerschen I've reviewed the diff and looks good. That said, Travis needs some love.

paulkerschen:
Thanks @johncrossman. Spooky interference with #201 resolved.

PR #203, opened 2016-11-10
COL-458 Make Canvas assignment categories invisible

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-458

Migrations to adjust database structure are already in place. This addition ensures that categories derived from assignments start out invisible.


johncrossman:
Looks good. Waiting on Travis.

PR #205, opened 2016-11-15
COL-416, allow PDFs and other assets to be downloaded

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-416

The `Download` button has been moved up, between `Edit` and `Delete`. The min number of buttons a user can see is one and the max is three - this is why bootstrap `col-xs` configs changed and `canManageAsset()` was move to control only the `Edit` button.

### DIFF
--- a/public/app/assetlibrary/item/item.html
+++ b/public/app/assetlibrary/item/item.html
@@ -4,15 +4,19 @@
 
   <!-- TITLE -->
   <div class="row assetlibrary-item-top">
-    <div class="col-xs-{{ asset.can_delete ? 5 : (canManageAsset() ? 8 : 12) }} col-sm-{{ asset.can_delete ? 7 : (canManageAsset() ? 10 : 12) }}">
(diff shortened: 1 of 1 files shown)


public/app/assetlibrary/item/item.html
```diff
@@ -26,7 +30,7 @@
     <div data-ng-if="asset.type !== 'whiteboard'">
 
       <!-- COMPLETED PREVIEWS -->
-      <div class="text-center assetlibrary-item-preview" id="assetlibrary-item-preview" data-ng-show="asset.preview_status === 'done'">
+      <div class="text-center assetlibrary-item-preview" id="assetlibrary-item-preview" data-ng-if="asset.preview_status === 'done'">
         <!-- Documents -->
```
johncrossman: **Note:** I changed `ng-show` to `ng-if` to get consistent with all other preview_status related logic in the template. These sections do not show/hide per background processes so why use `ng-show`?

paulkerschen:
Free your mind and your asset will download. Thanks @johncrossman.

PR #207, opened 2016-11-15
COL-536, developer docs (and related scripts) need to be consistent and easy to find 

johncrossman:
A dev's workstation should have Apache DOC_ROOT set to `~/Projects/collabosphere/public`. Therefore, this script should neither copy static files nor require env variable DOCUMENT_ROOT.


paulkerschen:
@johncrossman This makes sense. Do we have / should we have documentation about the recommended Apache DOC_ROOT setting?

johncrossman:
@paulkerschen Yes, we should. Do not merge until I document or find such doc.

johncrossman:
I've updated the README.md with links to files under `docs/`, renamed PDF to remove space and rebranded to SuiteC, where necessary. cc: @paulkerschen

johncrossman:
Fyi, you can see the README rendered at https://github.com/johncrossman/collabosphere/tree/NOJIRA_install-locally

paulkerschen:
Thanks @johncrossman, looks good.

PR #208, opened 2016-11-15
COL-473 Track discussion replies by reply ID, not parent entry ID

paulkerschen:
A fix for the reopened https://jira.ets.berkeley.edu/jira/browse/COL-473.


johncrossman:
Parent IDs just don't understand. Looks good, @paulkerschen

PR #213, opened 2016-11-16
COL-505 Do not award points for assets not included in library

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-505


johncrossman:
Looks right to me, @paulkerschen

PR #214, opened 2016-11-16
COL-538, we need not 'get latest code' when deploying to network

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-538

I noticed this extraneous instruction when reviewing release-instructions doc.


paulkerschen:
Very true.

PR #215, opened 2016-11-18
COL-541 Lower log level on discussion replies with missing parents

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-541

Since orphaned replies will call this every time the poller runs, a big grumpy error log is too noisy.

### DIFF
--- a/node_modules/col-canvas/lib/poller.js
+++ b/node_modules/col-canvas/lib/poller.js
@@ -909,10 +909,13 @@ var handleDiscussionEntryReply = function(ctx, activities, users, discussion, en
   // Find parent entry or reply
   var parent = (reply.parent_id === entry.id) ? entry : _.find(entry.recent_replies, {'id': reply.parent_id});
   if (!parent) {
-    log.error({
-      'replyId': reply.id,
+    log.debug({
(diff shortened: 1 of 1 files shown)


node_modules/col-canvas/lib/poller.js
```diff
@@ -909,10 +909,10 @@ var handleDiscussionEntryReply = function(ctx, activities, users, discussion, en
   // Find parent entry or reply
   var parent = (reply.parent_id === entry.id) ? entry : _.find(entry.recent_replies, {'id': reply.parent_id});
   if (!parent) {
-    log.error({
-      'replyId': reply.id,
+    log.debug({
       'discussionId': discussion.id,
-      'entry': entry,
+      'entryId': entry.id,
+      'replyId': reply.id
```
johncrossman: @paulkerschen  would `reply.parent_id` be useful in this log statement?

paulkerschen: Good thought, @johncrossman, added.

johncrossman:
Looks good, @paulkerschen

PR #217, opened 2016-11-18
COL-530 Disable email by default

paulkerschen:
Also, honor the alphabet.

https://jira.ets.berkeley.edu/jira/browse/COL-530


johncrossman:
It's an honor.

PR #218, opened 2016-11-18
COL-523 Reload asset state after editing details

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-523

I'm no ui-router expert but this seems to do the job.

http://stackoverflow.com/questions/21714655/reloading-current-state-refresh-data


johncrossman:
`merge: true`

PR #219, opened 2016-11-18
COL-551, asset downloads (assignment submissions) are branded 'suitec'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-551

A few hours of study on assignment submissions, the poller code and file naming leads to a one-word rebranding.


paulkerschen:
Looks like a million bucks to me.

PR #220, opened 2016-11-21
COL-516, roles are course-specific; full URNs in LTI spec describe context

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-516

Roles are recorded using full URN as seen in [LTI implementation-guide](https://www.imsglobal.org/specs/ltiv1p0/implementation-guide). Using namespaces avoids confusion with, for example, the `Instructor` role which is applicable in two contexts: institutional and course-level.

SQL script to migrate user data is included.


paulkerschen:
Thanks @johncrossman! Looks good.

PR #222, opened 2016-11-22
COL-497 Default created_at and updated_at values in Canvas table

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-497

See http://www.revsys.com/blog/2006/aug/04/automatically-updating-a-timestamp-column-in-postgresql/


johncrossman:
A very welcome enhancement. Thanks, @paulkerschen  Do you plan a similar trigger for all tables?

paulkerschen:
@johncrossman - My understanding is that Sequelize usually has its own logic for handling timestamps in the course of business, so database-level triggers might duplicate or conflict in other tables. The canvas table is an outlier in that we make changes directly in SQL rather than by way of Node.

johncrossman:
All makes sense. I'm sold.

PR #226, opened 2016-11-23
COL-516, Teacher role matches with LTI instructor full URN and short-name; ContentDeveloper, too

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-516

We are rolling forward on PR #220 by supporting both deprecated and full URN role names.


paulkerschen:
+1

PR #227, opened 2016-11-23
COL-534 Clarify that synced assignments appear as categories

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-534


johncrossman:
Yup, clarified. Thx @paulkerschen

PR #233, opened 2016-11-29
COL-480 Do not increment view count when creator views asset

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-480


johncrossman:
+1

PR #236, opened 2016-12-01
NOJIRA, remove start.sh from deploy.sh; post-deploy tasks might require downtime

johncrossman:
If there are no post-deploy tasks then Ops can run:
```
./deploy/deploy.sh -t 1.63 && ./deploy/start.sh
```
we must `exit 1` when `echo_usage` such that *start.sh* won't run in example above.


paulkerschen:
Looks good @johncrossman; giving Travis his say.

PR #239, opened 2016-12-02
COL-528 Make asset migration accessible to admins from the browser

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-528

This large PR is split into three commits for more manageable review.

The first commit moves asset migration logic into a module which is accessible both from the existing script and from a REST endpoint. Validation, sanity checks and tests are added.

The second commit creates a REST endpoint to list all courses associated with a user's Canvas account. We need this in order to display asset migration options to the user. The course listing includes filters relevant to this use case: 1) show only courses where I am an admin; 2) show only courses that are using the Asset Library; 3) don't show the current course.

The third commit includes front-end changes. The "Manage Categories" page becomes a "Manage Assets" page. The asset migration interface is added below the category options.

### DIFF
--- a/scripts/migrate_assets.js
+++ b/scripts/migrate_assets.js
@@ -25,56 +25,24 @@
  * ENHANCEMENTS, OR MODIFICATIONS.
  */
 
-var _ = require('lodash');
-var async = require('async');
-var contentDisposition = require('content-disposition');
-var fs = require('fs');
-var os = require('os');
-var fs = require('fs');
-var path = require('path');
-var request = require('request');
 var argv = require('yargs')
-    .usage('Usage: $0 --fromcourse [fromcourse] --fromuser [fromuser] --tocourse [tocourse] --touser [touser] --categories [categories]')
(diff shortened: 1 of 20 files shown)


scripts/migrate_assets.js
```diff
@@ -25,56 +25,24 @@
...
-var async = require('async');
-var contentDisposition = require('content-disposition');
-var fs = require('fs');
-var os = require('os');
-var fs = require('fs');
-var path = require('path');
-var request = require('request');
 var argv = require('yargs')
-    .usage('Usage: $0 --fromcourse [fromcourse] --fromuser [fromuser] --tocourse [tocourse] --touser [touser] --categories [categories]')
-    .demand(['fromcourse', 'fromuser', 'tocourse', 'touser'])
```
paulkerschen: Because a SuiteC user account is associated with only one course, the `fromcourse` and `tocourse` arguments are redundant.

johncrossman:
Very nice work, @paulkerschen  Commence merge procedure...

PR #240, opened 2016-12-05
COL-587 List users with case-insensitive ordering

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-587

This fixes a frequent test failure. The test expects case-insensitive ordering; code was performing a case-sensitive ordering; whether or not they matched depended entirely on the random strings.


johncrossman:
+1

PR #244, opened 2016-12-08
COL-524 Disable Canvas polling for a course after 90 days of inactivity; admins can re-enable

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-524


johncrossman:
Muy bueno. Thanks @paulkerschen

PR #245, opened 2016-12-09
COL-592 Deduplicate assignment submissions in Asset Library

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-592


johncrossman:
+1

PR #247, opened 2016-12-13
NOJIRA, prod suitec-ops will be updated deliberately, not automatically

johncrossman:
Pull script enhancements in prod _after_ they're road-tested in dev & qa.


paulkerschen:
Makes sense; superstitiously waiting on T.

PR #249, opened 2016-12-16
COL-607 Make poller's course deactivation threshold configurable

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-607

### DIFF
--- a/node_modules/col-canvas/lib/poller.js
+++ b/node_modules/col-canvas/lib/poller.js
@@ -49,7 +49,7 @@ var pollingDelay = null;
 var lastCoursePoll = Date.now();
 
 // The number of days a course may remain inactive before being removed from the poller's cycle
-var INACTIVE_DAYS_CUTOFF = 90;
+var deactivationThreshold = null;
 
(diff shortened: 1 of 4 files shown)


node_modules/col-canvas/lib/poller.js
```diff
@@ -204,8 +214,11 @@ var handleCourse = module.exports.handleCourse = function(course, opts, callback
           }
 
           // If the course has activity and that activity is in the distant past, we continue to the deactivation step; otherwise return.
```
johncrossman: @paulkerschen nit-picky question: The comment above ("...otherwise return") describes the return at line 222. Maybe move the comment down to avoid confusion with line 218?

johncrossman:
Looks good, @paulkerschen  Posted one question

node_modules/col-canvas/lib/poller.js
paulkerschen: @johncrossman Oh, yes, that's the right place for it. Moved!

johncrossman:
It slices, it dices! Activating deactivation threshold.

PR #250, opened 2016-12-17
COL-608 Allow asset migration to complete in background

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-608

Avoid timeouts while waiting for asset migration to complete.


johncrossman:
+1

PR #257, opened 2017-01-18
COL-631 Add health check for whiteboard thumbnail generation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-631


johncrossman:
Healthy whiteboards is what we want. Thx, @paulkerschen

PR #260, opened 2017-02-01
COL-616 Create .ebextensions for Elastic Beanstalk configuration

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-616

PR #263, opened 2017-02-01
[eb-master] COL-616 Create .ebextensions for Elastic Beanstalk configuration

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-616

eb-master version of #260.


johncrossman:
As before...

PR #264, opened 2017-02-01
COL-511 Require secure Postgres connection in production mode

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-511

This PR is for `eb-master` only.


johncrossman:
Secure connections are what the American people want. Thx, @paulkerschen

PR #265, opened 2017-02-13
COL-568 Properly format asset titles for screen reader

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-568

Screen reader wants its whitespace. Also use `div` for block-level elements, instead of inline elements with `display: block` styling.


johncrossman:
Screen readers be persnickety.  Thx, @paulkerschen

PR #269, opened 2017-02-14
COL-566 Primary button should precede Cancel button in tabbing sequence

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-566

This pattern is the same as in our bCourses tools (see https://github.com/ets-berkeley-edu/calcentral/pull/3941): put the primary button first in the DOM and order buttons right to left in CSS.


johncrossman:
Sequence is not always as it appears. Thx, @paulkerschen

PR #270, opened 2017-02-15
COL-646 Update user's last_activity on poller reactivation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-646

Updating this timestamp ensures that poller reactivation will stick even if there's been no other activity logged in the meantime.


johncrossman:
A finicky poller, indeed. Thx, @paulkerschen

PR #271, opened 2017-02-16
COL-435 COL-628 Exclude non-members from online user count; fix whiteboard paging bug

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-435

A side effect of this work was to diagnose and fix the mysterious
https://jira.ets.berkeley.edu/jira/browse/COL-583
https://jira.ets.berkeley.edu/jira/browse/COL-628


johncrossman:
Billions of non-members cost us the popular vote! Fake news!! (Thx, @paulkerschen)

PR #274, opened 2017-02-17
COL-650 Screen out hidden assets from daily/weekly emails

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-650


johncrossman:
I'll merge when green...

PR #276, opened 2017-02-17
COL-662 Remove inequality

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-662


johncrossman:
What can I say? I missed this in older PR. I messed up bigly.

image

PR #281, opened 2017-02-21
COL-657 Define editing:exited event before initializing whiteboard canvas

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-657

This PR is a matter of rearrangement within the large `whiteboardsBoardDirective.js`. The salient thing is that the `editing:exited` event must be added to the IText prototype _before_ we call `initializeCanvas()`; otherwise, IText objects created within that method won't inherit the event.


johncrossman:
+1

PR #282, opened 2017-02-21
[QA] COL-657 Define editing:exited event before initializing whiteboard canvas

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-657

QA version of #281.


johncrossman:
Waiting on T

PR #285, opened 2017-02-27
COL-655 Introduce configurable timeouts for background jobs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-655

Some SuiteC courses are now large enough that polling a single course takes longer than our five-minute timeout, which is raising false alarms. Bump the timeout to ten minutes and make it configurable.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -15,7 +15,8 @@
   "canvasPoller": {
     "enabled": true,
     "delay": 15,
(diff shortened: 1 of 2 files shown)


johncrossman:
Patience is bitter but its fruit is sweet.  Thx, @paulkerschen

PR #290, opened 2017-03-01
COL-535 Omit instructor assets from top asset category in weekly email

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-535

Regrettably, the weekly email still lacks a test suite.


johncrossman:
A legit omit. Thx, @paulkerschen

PR #293, opened 2017-03-06
COL-676 Don't show video player for undefined video_url

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-676


johncrossman:
+1

Waitin' on T.

PR #296, opened 2017-03-08
COL-462 Add UC copyright language

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-462

As discussed at the ticket, the infinite scroll in the asset and whiteboard lists means that the notice will get pushed down fairly quickly on those pages, but no one can deny it's there.

### DIFF
--- a/public/app/shared/copyright.html
+++ b/public/app/shared/copyright.html
@@ -1,0 +1,3 @@
+<div class="col-footer">
+  Copyright Notice and Disclaimer &mdash; Certain materials incorporated herein are Copyright &copy; 2016-2017 by The Regents of the University of California. All Rights Reserved.
+</div>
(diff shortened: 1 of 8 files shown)


public/app/shared/copyright.html
```diff
@@ -0,0 +1,3 @@
+<div class="col-footer">
+  Copyright Notice and Disclaimer – Certain materials incorporated herein are Copyright © 2016-2017 by The Regents of the University of California. All Rights Reserved.
```
johncrossman: @paulkerschen is "Disclaimer" followed by an em-dash?  Is `&mdash;` preferred?

paulkerschen: @johncrossman Entitified!

PR #297, opened 2017-03-09
COL-462 Place copyright notice within wrapper divs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-462

Follow-up to #296: some fanciness with wrapper divs and `ui-view` was causing the notice to appear twice.


johncrossman:
Sneaky. Thx, @paulkerschen

PR #298, opened 2017-03-09
COL-680 Improve messaging on whiteboard export failure

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-680

If whiteboard export is blocked by element preview status, show a more helpful message on both PNG downloads and Asset Library exports.


johncrossman:
I do love good messaging. Thx, @paulkerschen

PR #299, opened 2017-03-09
COL-508 Configurable timezone setting for app and logs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-508

Neither of the levels in our logging stack, https://www.npmjs.com/package/bunyan or
https://www.npmjs.com/package/bunyan-prettystream (whose developer says, "This library is only really meant for development and should not be used on production environments"), allow for timestamp formatting as such. However, we can elbow our way in by wrapping the individual logger functions.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -47,6 +47,7 @@
     "url": "http://localhost:2001/process",
     "apiKey": "some shared secret"
   },
+  "timezone": "America/Los_Angeles",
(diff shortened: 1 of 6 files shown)


johncrossman:
Bunyan says "don't use in prod". Peshaw! Thx, @paulkerschen

PR #300, opened 2017-03-15
COL-675 Do not sync assignment submissions unless sync explicitly enabled

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-675

The sheer number of to-be-deleted submission files in production means that a separate cleanup script isn't practical. Instead the poller will have to clean up its act as it goes along, requiring changes to the finicky process of submission syncing and tracking.

### DIFF
--- a/node_modules/col-canvas/lib/poller.js
+++ b/node_modules/col-canvas/lib/poller.js
@@ -37,6 +37,7 @@ var AssetsAPI = require('col-assets');
 var CategoriesAPI = require('col-categories');
 var CollabosphereConstants = require('col-core/lib/constants');
 var CourseAPI = require('col-course');
+var DB = require('col-core/lib/db');
 var log = require('col-core/lib/logger')('col-canvas/poller');
 var UsersAPI = require('col-users');
 
@@ -482,8 +483,7 @@ var handleAssignment = function(ctx, users, categories, opts, assignment, callba
 };
 
 /**
- * Handle the category synchronisation for an assignment. If a new category
- * gets creates it will get added to the set of categories
+ * Handle the category synchronisation for an assignment, creating a new category if necessary
  *
  * @param  {Context}          ctx                                   Standard context containing the current user and the current course
  * @param  {Object}           users                                 The users in the course indexed by their `canvas_user_id` attribute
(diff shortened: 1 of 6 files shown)


node_modules/col-canvas/lib/poller.js
```diff
@@ -482,8 +483,7 @@ var handleAssignment = function(ctx, users, categories, opts, assignment, callba
 };
 
 /**
- * Handle the category synchronisation for an assignment. If a new category
- * gets creates it will get added to the set of categories
+ * Handle the category synchronisation for an assignment, creating a new category if necessary
```
johncrossman: @paulkerschen I like that you preserved the British spelling of "synchronisation".

node_modules/col-canvas/lib/poller.js
```diff
@@ -626,106 +626,133 @@ var handleSubmission = function(ctx, users, assignment, activities, submissionAs
     'objectType': CollabosphereConstants.ACTIVITY.OBJECT_TYPES.CANVAS_SUBMISSION,
     'metadata': {
       'submission_id': submission.id,
-      'attempt': submission.attempt
+      'attempt': submission.attempt,
+      'sync_files': assignment.category.visible
```
johncrossman: @paulkerschen Nit-picky comment: If 'sync_files' is a boolean I would expect the key name to be an adjective (or similar), not a noun. E.g., 'requires_sync'. Am I understanding its purpose?

johncrossman: Note: I understand that 'sync_files' is meant as an action description. But lazy devs (like me) might not immediately recognize that intention.

johncrossman:
Although it was authored in the wee hours of the night it looks good. Thx, @paulkerschen  I posted one question.

node_modules/col-canvas/lib/poller.js
paulkerschen: Thanks for attending to the parts of speech, @johncrossman. Key has been changed to `file_sync_enabled`.

johncrossman:
Thus we make the Great Leap Forward.

PR #303, opened 2017-03-15
NOJIRA, update .gitignore with: apache/suitec.conf

johncrossman:
An overdue fix. PDG devs will need to regenerate Apache config locally to pick up new LTI tool (dashboard). We don't want the generated `suitec.conf` to end up in Git repo.

cc: @paulkerschen @sandeepmjay

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 *.log
 logs/*.log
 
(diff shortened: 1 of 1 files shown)


paulkerschen:
+1

PR #304, opened 2017-03-20
COL-714 Avoid crash on errored whiteboard thumbnails

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-714

Note that this PR is made against a `1.10.1-patch` branch, as the `qa` branch is already queued up for 1.11.


johncrossman:
That feeling when a faulty log statement forces a patch deploy...  Thanks, @paulkerschen

PR #305, opened 2017-03-22
COL-712, MyActivity section on Dashboard splash for demo

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-712

In prep for meeting with UX peeps, it would be nice to have something to get conversation started.  These pieces will be valuable when we plug in more (real) data. Header of splash page is my-activity include.

image

The files under `img/mockup` will be removed once we have UX.


paulkerschen:
Thanks @johncrossman; looks like a good place to start.

PR #309, opened 2017-03-29
COL-737, enable linking to user profile by id (default: me)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-737

* change to `app.bootstrap.js` will parse out GET params from window.location
* new `/profile/:id` path takes advantage of `id_parameter` made available by change above
* I have put off the renaming of files, paths.  We'll refactor once official name is in.

### DIFF
--- a/node_modules/col-users/lib/rest.js
+++ b/node_modules/col-users/lib/rest.js
@@ -41,6 +41,19 @@ Collabosphere.apiRouter.get('/users/me', function(req, res) {
 });
 
 /*!
+ * Get user of current course by id
+ */
+Collabosphere.apiRouter.get('/users/id/:id', function(req, res) {
+  UsersAPI.getUsers(req.ctx, [req.params.id], function(err, users) {
+    if (err) {
+      return res.status(err.code).send(err.msg);
+    }
+
+    return res.status(200).send(users[0]);
+  });
+});
+
+/*!
  * Get all users in the current course
  */
 Collabosphere.apiRouter.get('/users', function(req, res) {
--- a/public/app/app.bootstrap.js
+++ b/public/app/app.bootstrap.js
@@ -55,9 +55,15 @@
    * @see https://css-tricks.com/snippets/jquery/get-query-params-object/
    */
   var getQueryParameters = function() {
-    return decodeURIComponent(document.location.search).replace(/(^\?)/,'').split('&').map(function(n) {
+    var queryArgs = decodeURIComponent(document.location.search).replace(/(^\?)/,'');
+    var parameters = queryArgs.split('&').map(function(n) {
       return n = n.split('='), this[n[0]] = n[1], this;
     }.bind({}))[0];
+
+    // Extract '_id' from embedded toolUrl to support deep-linking from other SuiteC tool or email, possible.
+    var m = queryArgs.match(/.*[\?&]_id=([0-9a-zA-Z]+).*/);
+    parameters.deep_link_id = (m && m.length > 0) ? m[1] : null;
+    return parameters;
   };
 
   /**
@@ -80,6 +86,7 @@
     }).then(function(results) {
       collabosphere.constant('me', results.me.data);
       collabosphere.constant('config', results.config.data);
+      collabosphere.constant('deepLinkId', parameters.deep_link_id);
     });
   };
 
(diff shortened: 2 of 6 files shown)


johncrossman:
I'll fix this shortly

node_modules/col-users/lib/rest.js
```diff
@@ -41,6 +41,26 @@ Collabosphere.apiRouter.get('/users/me', function(req, res) {
...
 /*!
+ * Get user of current course by id
+ */
+Collabosphere.apiRouter.get('/users/:id', function(req, res) {
+  // Get user by id
+  var userId = req.params.id;
+  // Only admins can access non-self user data
+  if (userId != req.ctx.user.id && !req.ctx.user.is_admin) {
+    return res.status(401).send('Unauthorized API request');
+  }
```
paulkerschen: @johncrossman Is it the case that this endpoint will be totally inaccessible to non-admins? It seems like we might want to return at least a subset of properties that users have made public.

Also, convention would be for authentication checks to take place in `api.js`, not `rest.js`. We might want a new UserAPI method that would retrieve the User object for ID, validate that the user exists in ctx.course (otherwise returning 400), and filter out any properties that should be hidden from non-admins.

public/app/app.bootstrap.js
```diff
@@ -55,7 +55,7 @@
    * @see https://css-tricks.com/snippets/jquery/get-query-params-object/
    */
   var getQueryParameters = function() {
-    return decodeURIComponent(document.location.search).replace(/(^\?)/,'').split('&').map(function(n) {
+    return decodeURIComponent(document.location.search).replace(/(^\?)/,'').split(/[&,?]/).map(function(n) {
```
paulkerschen: @johncrossman So after we've stripped out the first question mark, another question mark might be present within the query string itself as a delimiter? Is it intentional to also include commas as delimiters?

public/app/app.bootstrap.js
```diff
@@ -80,6 +80,10 @@
     }).then(function(results) {
       collabosphere.constant('me', results.me.data);
       collabosphere.constant('config', results.config.data);
+      // Support deep linking with generic 'by_id' param passed from other SuiteC tool or email.
+      if (parameters.by_id) {
+        collabosphere.constant('id_parameter', parameters.by_id);
+      }
```
paulkerschen: @johncrossman Would jscs tolerate `parameters['_id']`?

paulkerschen: @johncrossman I hadn't dug into this `collabosphere.constant` stuff before, but it seems like a clever way to cache the id parameter. Might there be an issue with this parameter hanging around longer than expected and interfering with subsequent navigation? If so perhaps we could explicitly clear it once it's done its job.

johncrossman: A good question/suggestion, @paulkerschen  I'll investigate.

node_modules/col-users/lib/rest.js
johncrossman: @paulkerschen Yes, I'm aware that we'll be showing other users a filtered view of the profile. I put this in as a placeholder for "some kind of restriction". I'll remove the placeholder – I'm sure we'll remember said requirement.

johncrossman:
@paulkerschen Amendments are in but I anticipate Travis unhappiness [similar to before](https://travis-ci.org/ets-berkeley-edu/suitec/builds/216159930). Investigating...

public/app/app.bootstrap.js
```diff
@@ -80,6 +86,7 @@
     }).then(function(results) {
       collabosphere.constant('me', results.me.data);
       collabosphere.constant('config', results.config.data);
+      collabosphere.constant('deepLinkId', parameters.deep_link_id);
```
johncrossman: @paulkerschen In my local testing, this block of code is hit with every request. I'd like to believe we have no risk persist the attribute across session. Is there an edge case here?

paulkerschen: Not that I know of, @johncrossman; if we're hitting this block of code every time and it's explicitly set to null when missing, I think we're good.

paulkerschen:
Happy Travis! Thanks @johncrossman.

PR #315, opened 2017-04-03
COL-746 Fix source tags in asset library detail view

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-746

See https://docs.angularjs.org/api/ng/directive/ngSrc


johncrossman:
Thanks, @paulkerschen  Waiting on Sir T.

PR #319, opened 2017-04-04
COL-737, link names on asset detail to profile if target tool is available

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-737

`crossLtiSupport` (i.e., linking between LTI tools) is now a shared utility. Future cross-LTI actions might include linking to assets from profile page.

### DIFF
--- a/public/app/shared/toolHrefDirective.js
+++ b/public/app/shared/toolHrefDirective.js
@@ -1,0 +1,73 @@
+/**
+ * Copyright ©2017. The Regents of the University of California (Regents). All Rights Reserved.
+ *
+ * Permission to use, copy, modify, and distribute this software and its documentation
+ * for educational, research, and not-for-profit purposes, without fee and without a
+ * signed licensing agreement, is hereby granted, provided that the above copyright
+ * notice, this paragraph and the following two paragraphs appear in all copies,
+ * modifications, and distributions.
+ *
+ * Contact The Office of Technology Licensing, UC Berkeley, 2150 Shattuck Avenue,
+ * Suite 510, Berkeley, CA 94720-1620, (510) 643-7201, otl@berkeley.edu,
+ * http://ipira.berkeley.edu/industry-info for commercial licensing opportunities.
+ *
+ * IN NO EVENT SHALL REGENTS BE LIABLE TO ANY PARTY FOR DIRECT, INDIRECT, SPECIAL,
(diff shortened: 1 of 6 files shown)


paulkerschen:
@johncrossman This makes sense, but two questions on naming:
1) Will this pattern be used for routing within the dashboard tool (e.g. navigating between profiles) as well as routing between LTI tools? If so, `crossLtiSupport` might be a misleading name.
2) `provider` seems like an opaque name for the current course, and I didn't quite follow why `crossLtiSupport` needs to return it if it's the same argument with `crossLtiSupport` is initialized. Is there any reason not to just use the `$scope.course` or `$scope.me.course` we have available?

johncrossman:
@paulkerschen 
1. No, it was not my intention to use this "support" to navigate _within_ the new tool. The hope is that other cross-LTI scenarios (e.g., linking to asset from profile) would use same `crossLtiSupport()`
2. I agree on the opaque naming – amendment coming. And yes to `$scope.me.course`! Who knew?!

public/app/shared/utilService.js
```diff
@@ -89,29 +89,6 @@
...
-     * @param  {Object}       course          Course with one or more SuiteC tools
-     * @param  {Object}       user            Member of course
-     * @return {String}                       URL with necessary query args
-     */
-    var getUserDashboardUrl = function(course, user) {
-      var dashboardUrl = getGenericDashboardUrl(course);
-      return dashboardUrl && user ? dashboardUrl + '?_id=' + user.id : null;
-    };
-
-    /**
```
johncrossman: @paulkerschen The above logic was moved to a directive. Appropriate? Better? I'll let you be the judge.

public/app/shared/toolHrefDirective.js
```diff
@@ -0,0 +1,73 @@
...
+              return course.whiteboards_url;
+            }
+            default: {
+              return '';
+            }
+          }
+        };
+        var queryArgs = scope.id ? '?_id=' + scope.id : '';
+        elem.attr('href', getToolUrl(scope.course, attrs.tool) + queryArgs);
+        elem.attr('target', '_parent');
```
johncrossman: Our new LTI tool will link to most/all other SuiteC tools.

paulkerschen:
I think the directive is sweet, @johncrossman. Thanks!

PR #320, opened 2017-04-05
COL-752 Script to update LTI configurations in Canvas

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-752

This repointing script is more complex than its Junction cousin since we're dealing with multiple Canvases and can't provide account IDs as a configuration option. To figure out where a given tool is configured, we may need to iterate up through parent accounts.

### DIFF
--- a/scripts/update_lti.js
+++ b/scripts/update_lti.js
@@ -1,0 +1,254 @@
+#!/usr/bin/env node
+
+/**
+ * Copyright ©2017. The Regents of the University of California (Regents). All Rights Reserved.
+ *
+ * Permission to use, copy, modify, and distribute this software and its documentation
+ * for educational, research, and not-for-profit purposes, without fee and without a
+ * signed licensing agreement, is hereby granted, provided that the above copyright
+ * notice, this paragraph and the following two paragraphs appear in all copies,
+ * modifications, and distributions.
+ *
+ * Contact The Office of Technology Licensing, UC Berkeley, 2150 Shattuck Avenue,
+ * Suite 510, Berkeley, CA 94720-1620, (510) 643-7201, otl@berkeley.edu,
+ * http://ipira.berkeley.edu/industry-info for commercial licensing opportunities.
+ *
+ * IN NO EVENT SHALL REGENTS BE LIABLE TO ANY PARTY FOR DIRECT, INDIRECT, SPECIAL,
+ * INCIDENTAL, OR CONSEQUENTIAL DAMAGES, INCLUDING LOST PROFITS, ARISING OUT OF
+ * THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN IF REGENTS HAS BEEN ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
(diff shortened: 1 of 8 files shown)


scripts/update_lti.js
```diff
@@ -0,0 +1,254 @@
...
+ *
+ * @param  {Canvas}           canvas              The Canvas object to update
+ * @param  {Number}           accountId           The id of the account to update
+ * @param  {Object[]}         toolsToUpdate       Tools to be updated 
+ * @param  {Function}         callback            Standard callback function
+ */
+var updateToolsForAccount = function(canvas, accountId, toolsToUpdate, callback) {
+  CanvasAPI.getExternalToolsForAccount(canvas, accountId, function(err, accountTools) {
+    if (err) {
+      log.error({'err': err, 'account': account.id, 'accountId': accountId}, 'Failed to get external tool configurations');
```
johncrossman: @paulkerschen Remove `'account': account.id`?

scripts/update_lti.js
```diff
@@ -0,0 +1,254 @@
...
+        if (err) {
+          log.error({'err': err, 'course': course.id, 'tool': tool}, 'Failed to update tool');
+          results.error.push(tool);
+        } else {
+          log.info({'tool': tool, 'course': course.id}, 'Updated tool at the course level');
+          results.success.push(tool);
+        }
+
+        // Whether we've succeeded or errored, we shouldn't try this tool again. Null out the array at this index.
+         toolsToUpdate[index] = null;
```
johncrossman: @paulkerschen a touch too much indentation

scripts/update_lti.js
```diff
@@ -0,0 +1,254 @@
...
+        if (err) {
+          log.error({'err': err, 'account': accountId}, 'Failed to get account properties');
+          results.error = results.error.concat(toolsToUpdate);
+          return callback();
+        }
+        if (!accountProperties.parent_account_id) {
+          // If we can't find a parent account ID, mark the remaining tools as not found.
+          results.notfound = results.notfound.concat(toolsToUpdate);
+          return callback();
+        }
```
johncrossman: @paulkerschen Do you see value in using `/api/v1/accounts/:account_id/roles` to verify `lti_add_edit` privilege prior to `updateToolsForAccount`? Ideally, we do such verification weeks before the AWS cutover.

johncrossman:
@paulkerschen My review is done – nice work! I've posted a few comments/questions, with only one must-fix.

scripts/update_lti.js
paulkerschen: @johncrossman Ah, I didn't know about that endpoint. I agree we should verify, but perhaps best done in an advance step separate from this script? I think confusing Canvas terminology means that "account" in the endpoint refers to a user account (e.g. the SuiteC SPA), whereas "account" in `updateToolsForAccount` refers to a subaccount (e.g. "Official Courses").

paulkerschen:
Thanks for review, @johncrossman, amendments made.

johncrossman:
You know how I love da' scripts! Thx, @paulkerschen

PR #324, opened 2017-04-10
COL-725, copyright notice to add/edit asset, add whiteboard

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-725

The only change in `dashboard/splash.html` is a wrapper div, with copyright notice at bottom.  Put `?w=1` for simpler diff.


paulkerschen:
Ayup.

PR #325, opened 2017-04-10
COL-769, when 'My Assets' is empty on profile then offer Upload / Add Link asset

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-769

UX Jira: https://jira.ets.berkeley.edu/jira/browse/COL-763

User on profile page, with no assets, can click to upload a new asset or learn more about bookmarklet. For these target pages to have "Back to Dashboard" we must extend the cross-tool-href paradigm by allowing `id` to carry a router-state name, as seen in `app.states.js`.

### DIFF
--- a/public/app/assetlibrary/item/item.html
+++ b/public/app/assetlibrary/item/item.html
@@ -1,7 +1,7 @@
 <div data-ng-show="state.current.name === 'assetlibrarylist.item'">
   <a ui-sref="assetlibrarylist($parent.searchOptions)" class="col-back" data-ng-click="backToAssetLibrary()" data-ng-if="!whiteboardReferral && !referringTool"><i class="fa fa-angle-left"></i> Back to Asset Library</a>
 
-  <a tool-href data-tool="dashboard" data-course="me.course" data-id="referringId" class="col-back" data-ng-if="referringTool === 'dashboard'"><i class="fa fa-angle-left"></i> Back to Dashboard</a>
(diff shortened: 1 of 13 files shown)


public/app/assetlibrary/item/item.html
```diff
@@ -1,7 +1,7 @@
 <div data-ng-show="state.current.name === 'assetlibrarylist.item'">
   <a ui-sref="assetlibrarylist($parent.searchOptions)" class="col-back" data-ng-click="backToAssetLibrary()" data-ng-if="!whiteboardReferral && !referringTool"><i class="fa fa-angle-left"></i> Back to Asset Library</a>
 
-  <a tool-href data-tool="dashboard" data-course="me.course" data-id="referringId" class="col-back" data-ng-if="referringTool === 'dashboard'"><i class="fa fa-angle-left"></i> Back to Dashboard</a>
+  <a tool-href data-tool="dashboard" data-course="me.course" data-id="referringTool.id" class="col-back" data-ng-if="referringTool.name === 'dashboard'"><i class="fa fa-angle-left"></i> Back to Dashboard</a>
```
johncrossman: *Note:* This asset item view cannot use `backtoassetlibrary.html` because the 'Back' link has custom logic w.r.t search filters.

paulkerschen:
Code makes sense, @johncrossman. One question: is `backtoassetlibrary` the best name for the template and controller, given that (if I'm reading correctly) the reason for the custom code is that we might be telling the link to lead somewhere other than the asset library? Perhaps just `backlink` or `backtoreferrer` would be clearer?

johncrossman:
@paulkerschen Yes, `backtoassetlibrary` does not tell the whole stroy. I'll do `backtoreferrer`.

johncrossman:
Amendments are in. Travis is chewing...

PR #328, opened 2017-04-11
COL-771, message: No matching assets; remove/refactor for consistency

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-771

In addition to the standard "no matching assets" message, I have done some renaming/removing to add some sense, despite the delayed naming of new LTI tool. The tool is "dashboard" and the view is "profile".

(Note: first attempt at this PR was #327 which had conflict)


public/app/dashboard/profile.html
```diff
@@ -1,27 +1,52 @@
...
+    <div class="profile-activity-timeline"
          data-activity-timeline="userAssetActivity"
          data-labels-width="200"></div>
   </div>
 
-  <div class="featured-assets-container" data-ng-if="isMyProfile || featuredAssets.length">
+  <div class="profile-featured-assets-container" data-ng-if="isMyProfile || featuredAssets.length">
     <h3>
-      <span class="featured-list-name" data-ng-if="!featuredAssets.length">My Assets</span>
+      <span data-ng-if="!featuredAssets.length">My Assets</span>
```
johncrossman: Removal of `featured-list-name` style removes boldness.

paulkerschen:
If names be not correct, language is not in accordance with the truth of things. If language be not in accordance with the truth of things, affairs cannot be carried on to success. When affairs cannot be carried on to success, proprieties and music do not flourish. When proprieties and music do not flourish, punishments will not be properly awarded. When punishments are not properly awarded, the people do not know how to move hand or foot. Therefore a superior man considers it necessary that the names he uses may be spoken appropriately, and also that what he speaks may be carried out appropriately.

johncrossman:
Word up.

PR #329, opened 2017-04-11
COL-772 Revert to eventdrops-compatible d3 version

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-772

Bower cache struck again, misleading me as to what major version of d3 I was using.


johncrossman:
PDG's love/hate relationship with caches goes on. Thx, @paulkerschen

PR #301, opened 2017-03-15
COL-706, Impact Studio LTI tool gets general purpose name: Dashboard

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-706
https://jira.ets.berkeley.edu/jira/browse/COL-707
https://jira.ets.berkeley.edu/jira/browse/COL-708

Introduction of the new LTI tool begins with proposed naming: Dashboard. Of course, the content of the "Impact Studio" landing page is up to UX / Service Lead. Nonetheless, I'd prefer to keep under-the-hood naming descriptive, generalized and NOT based on transient grant initiatives.

Before I proceed with such naming I use this PR to get approval from da team.

PR #330, opened 2017-04-11
COL-708 Add dashboard RewriteRule to ebextensions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-708

Follow-up to #301.


johncrossman:
Tool_4 is open for business. Thx, @paulkerschen

PR #331, opened 2017-04-11
COL-756, get assets by user id for My Assets

johncrossman:
Such are the hazards of testing locally, with a single user id. Bug fixed.


paulkerschen:
Indeed!

PR #333, opened 2017-04-11
COL-762 Prototype per-asset activity timeline

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-762


johncrossman:
Thanks much, @paulkerschen

PR #334, opened 2017-04-11
COL-759, profile page > sortAssets() after getUser is done

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-759

Null user is a problem.


paulkerschen:
Right.

PR #336, opened 2017-04-12
COL-704, updated community asset filters: Trending, Pinned, etc

johncrossman:
UX: https://jira.ets.berkeley.edu/jira/browse/COL-704

@paulkerschen Swimlane filter links did not have proper mouseover pointer. What's the right style fix here w.r.t `href`?


paulkerschen:
@johncrossman A quick search and check of the Angular docs suggests that empty`href` is ok, and surely preferable to getting CSS involved.

PR #337, opened 2017-04-12
COL-782 Include error details in logger output

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-782

This PR is a workaround pending a better overall log setup.


johncrossman:
From the makers of Godzilla vs. Mothra... Grotesque workaround vs. Bunyan!!  Frighteningly effective, @paulkerschen

PR #340, opened 2017-04-12
NOJIRA Fix to LTI repointing script

paulkerschen:
A small but important fix to ensure the LTI script is grabbing a real tool ID instead of a surprising number from a surprising URL.


johncrossman:
Love for the regex.

PR #342, opened 2017-04-13
COL-780 Scale down large whiteboards on export

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-780

Tested out on col-qa.


johncrossman:
These regulations will kill innovation!  Nonetheless... Well done, @paulkerschen

PR #344, opened 2017-04-13
COL-789, include hash in hashtag pattern match

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-789

`assetLibraryService` expects leading `#` char. Including it in pattern match will send user to listing page, as expected.


paulkerschen:
Sneaky, sneaky capture groups!

PR #349, opened 2017-04-15
COL-745, JSCS is deprecated; move to ESLint

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-745

Rules on `@return` account for most of the diff. Nothing objectionable, I hope. In the eslint YAML file we have:
* enabled rules set to 2 
* disabled rules set to 0
* rules want to enable commented out (because code needs work) 

More info:
* http://jscs.info/
* https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2


paulkerschen:
ESLint really wants us to contemplate the void, I guess. Thanks @johncrossman

PR #352, opened 2017-04-20
COL-746 Use ng-attr to avert early fetch attempt on video poster image

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-746

#315 didn't fix the problem. It's always in the last place you look.


johncrossman:
This time, for sure! Thanks, @paulkerschen

PR #355, opened 2017-04-24
COL-794, aggregatedSearch: findAssets added to front-end controller

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-794

`$scope.search` properties are structured according to the complexity of aggregatedSearch page. You'll notice cautious code-sharing with `assetLibraryListController`.

### DIFF
--- a/public/app/dashboard/profileController.js
+++ b/public/app/dashboard/profileController.js
@@ -134,9 +134,6 @@
     };
 
     var init = function() {
-      // Advanced search is not supported on profile page
-      $scope.supportsAdvancedSearch = false;
-
       // Determine user
       if (deepLinkId) {
         userFactory.getUser(deepLinkId).success(function(user) {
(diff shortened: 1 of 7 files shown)


public/app/dashboard/profileController.js
```diff
@@ -134,9 +134,6 @@
     };
 
     var init = function() {
-      // Advanced search is not supported on profile page
-      $scope.supportsAdvancedSearch = false;
-
```
johncrossman: False is default – no need to be explicit.

public/app/dashboard/search/results.html
```diff
@@ -1,15 +1,50 @@
...
+                  <div class="col-threedots">
+                    <small>by {{asset.users.length > 1 ? asset.users.length + " collaborators" : asset.users[0].canvas_full_name}}</small>
+                  </div>
+                </div>
+              </div>
+              <div class="text-right col-list-item-actions" iconbar asset="asset" viewonly="true"></div>
+            </div>
+          </a>
+        </li>
+      </ul>
```
johncrossman: The UX above will change. For now, we have `<ul>` as seen in assetLibrary list view.

paulkerschen:
Thanks @johncrossman.

PR #357, opened 2017-04-25
COL-788 Replace activity timeline scroll zoom with buttons

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-788

Buttons are not yet properly styled or positioned, but are hooked up to a programmatic zoom in place of gesture scrolling.
 
Indefinite zoom-out is also disabled (https://jira.ets.berkeley.edu/jira/browse/COL-777).


bower.json
```diff
@@ -21,8 +21,8 @@
     "angular-strap": "2.3.10",
     "angular-ui-router": "0.2.18",
     "bootstrap": "3.3.7",
-    "d3": "3.5.17",
-    "event-drops": "marmelab/eventDrops#08210231b851854ae3526e54ebfe04fc0b0f1092",
+    "d3": "4.8.0",
+    "event-drops": "kwesterfeld/eventDrops#e39b5fd108ec5d182003ea4bc5f063926e30e310",
```
paulkerschen: Event-drops has no official release with the features we need (another reason we may eventually need to fork). For the moment, this points to a revision that works for us. Much of the legwork in this PR is a matter of build settings for the latest xtra-fancy event-drops and d3.

johncrossman:
Makes sense to me, @paulkerschen  On a related note, why hasn't the eventDrops project – which is fairly active in commits – cut an official release since Jun 2015?!

PR #358, opened 2017-04-25
COL-799 Use 24-hour format in activity timelines

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-799


johncrossman:
Merged at 15:25.  Thx, @paulkerschen

PR #360, opened 2017-04-26
COL-839, profile search prototype: User-only w/ autosuggest keeps you on profile page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-839

Search is hard, so let's start with something simple. Put focus on search and you get list of all course users. Start typing and list narrows. Click or hit return on a user and page reloads with selected profile.  See screenshot.

Perhaps we'll end up with a more sophisticated search but... One step at a time.

image

### DIFF
--- a/public/app/app.states.js
+++ b/public/app/app.states.js
@@ -81,7 +81,7 @@
         'templateUrl': '/app/dashboard/profile.html',
         'controller': 'ProfileController'
       })
-      .state('dashboard.profile.user', {
+      .state('userprofile', {
         'url': '/profile/:userId',
         'templateUrl': '/app/dashboard/profile.html',
         'controller': 'ProfileController'
--- a/public/app/dashboard/search/profileSearchAutosuggestFilter.js
+++ b/public/app/dashboard/search/profileSearchAutosuggestFilter.js
@@ -1,0 +1,38 @@
+/**
+ * Copyright ©2017. The Regents of the University of California (Regents). All Rights Reserved.
+ *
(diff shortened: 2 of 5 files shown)


public/app/app.states.js
```diff
@@ -81,7 +81,7 @@
         'templateUrl': '/app/dashboard/profile.html',
         'controller': 'ProfileController'
       })
-      .state('dashboard.profile.user', {
+      .state('userprofile', {
```
johncrossman: Oddly, `$state.go(...)` in profileController was unhappy with dots in state name (above). So, I changed the state name. Duh!

public/app/dashboard/search/profileSearchAutosuggestFilter.js
```diff
@@ -0,0 +1,38 @@
...
+
+  'use strict';
+
+  angular.module('collabosphere').filter('userProfileSearchDropdown', ['$sce', function($sce) {
+    return function(label, query, option) {
+      var html = option.is_admin ? '<i class="fa fa-graduation-cap"></i> ' + label : label;
+
+      return $sce.trustAsHtml(html);
+    };
+  }]);
```
johncrossman: The slightly cumbersome code above (1) adheres to `oi-select` pattern and (2) puts a (rightfully deserved?) `graduation-cap` on instructors, as seen elsewhere.

paulkerschen:
Oi! Select! Thanks @johncrossman

PR #361, opened 2017-04-26
COL-828 Add 'impact' and 'trending' columns to assets

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-828

### DIFF
--- a/node_modules/col-core/lib/db.js
+++ b/node_modules/col-core/lib/db.js
@@ -468,6 +468,26 @@ var setUpModel = function(sequelize) {
       'type': Sequelize.INTEGER,
       'defaultValue': 0
     },
+    'impact_percentile': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false,
+      'defaultValue': 0
+    },
+    'impact_score': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false,
+      'defaultValue': 0
+    },
+    'trending_percentile': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false,
+      'defaultValue': 0
+    },
+    'trending_score': {
+      'type': Sequelize.INTEGER,
+      'allowNull': false,
+      'defaultValue': 0
+    },
     'visible': {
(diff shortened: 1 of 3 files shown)


node_modules/col-core/lib/db.js
```diff
@@ -468,6 +468,22 @@ var setUpModel = function(sequelize) {
...
+      'defaultValue': 0
+    },
+    'trending_percentile': {
+      'type': Sequelize.INTEGER,
+      'defaultValue': 0
+    },
+    'trending_score': {
+      'type': Sequelize.INTEGER,
+      'defaultValue': 0
+    },
```
johncrossman: @paulkerschen no need for `'allowNull': false` above?

paulkerschen: @johncrossman Good question; existing code is inconsistent on that point. But I've amended the new columns to make it explicit.

scripts/20170426-col828/add_impact_trending_to_assets.sql
```diff
@@ -0,0 +1,15 @@
+-- Add per-asset columns containing 'impact' and 'trending' scores and percentiles. These values
+-- are periodically recalculated and used interally for query selection and ordering, but do not 
+-- surface through the API.
+
+ALTER TABLE assets ADD impact_percentile integer NOT NULL DEFAULT 0;
+ALTER TABLE assets ADD impact_score integer NOT NULL DEFAULT 0;
+ALTER TABLE assets ADD trending_percentile integer NOT NULL DEFAULT 0;
+ALTER TABLE assets ADD trending_score integer NOT NULL DEFAULT 0;
```
johncrossman: @paulkerschen Fyi, I believe the four ADDs above can live in a single ALTER statement:
```
ALTER TABLE table ADD COLUMN col1 int, ADD COLUMN col2 int, ...;
```
Gives you benefit of all or nothing. But, this is a one time script so I don't see a need for change.

johncrossman:
Waiting on T...

PR #363, opened 2017-04-26
COL-839, loadOtherUsers() after loadProfile(); hide oi-select arrow

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-839

Two small adjustments:
* Because we're not using `multiple` in `oi-select` we get an unwanted arrow in input element. This is not the first oi-select customization in our code...
* The previous placement of `$scope.otherUsers` assignment could have been a problem. Let's avoid the risk.

### DIFF
--- a/public/app/dashboard/profileController.js
+++ b/public/app/dashboard/profileController.js
@@ -139,24 +139,26 @@
       $scope.user.hashtags = ['#badminton', '#bridge', '#break-dancing'];
     };
 
+    var loadOtherUsers = function(user) {
+      // The dropdown of 'search for other users' needs the following
+      // list of otherUsers. We exclude user.id of current profile view.
+      userFactory.getAllUsers().then(function(response) {
+        $scope.otherUsers = _.reject(response.data, {'id': user.id});
+      });
+    };
+
     var init = function() {
       // Determine user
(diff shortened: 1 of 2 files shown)


public/app/dashboard/profileController.js
```diff
@@ -139,24 +139,28 @@
       $scope.user.hashtags = ['#badminton', '#bridge', '#break-dancing'];
     };
 
+    var loadOtherUsers = function(user) {
+      // The dropdown of 'search for other users' needs the following
+      // list of otherUsers. We exclude user.id of current profile view.
+      userFactory.getAllUsers().then(function(response) {
+        $scope.otherUsers = _.reject(response.data, function(otherUser) {
```
paulkerschen: @johncrossman I believe Lodash supports a one-liner shorthand here:  
```
_.reject(response.data, {'id': user.id});
```

johncrossman: Nifty! Amending now.

johncrossman: The Lodash shorthand was tested and works. Amendment is in.

paulkerschen:
Waiting for T to clean out the lint...

PR #364, opened 2017-04-26
COL-824 Create get_whiteboard_add_asset activities for asset creators

paulkerschen:
Note that this logic, as with so much having to do with whiteboards, lives behind the barrier of websockets and is inaccessible to our testing framework.

https://jira.ets.berkeley.edu/jira/browse/COL-824


johncrossman:
The mystique of websockets... Thx, @paulkerschen

PR #365, opened 2017-04-27
COL-837 Add remix_whiteboard and get_remix_whiteboard activities

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-837


johncrossman:
Thx, @paulkerschen, for puttin' it in the mix...

image

PR #366, opened 2017-04-27
COL-829 Add impact to activity type configuration

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-829

Secret-sauce values from the JIRA are multiplied by 2 to keep everything an integer.

Mentions, cites, and pins are not yet implemeted.


johncrossman:
It all adds up. Waiting on T.

PR #371, opened 2017-05-01
COL-846 Do not pass null email to users API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-846

Superstrict validation in col-users/lib/api.js says that undefined email is OK, null email is not. Make the poller honor the contract.


johncrossman:
Thx, @paulkerschen

PR #379, opened 2017-05-02
COL-804, only show 'Everyone's Assets' swimlane when user is on his/her own profile

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-804

As a wise man once said, we are not building a view-as feature.


paulkerschen:
If someone visits _my_ page to view _everyone's_ assets, am _I_ impactful? I think not. Thanks @johncrossman

PR #380, opened 2017-05-03
COL-784, only show profile swimlanes when isLoading=false

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-784

Added `isLoading` for user and community swimlanes. Plus, added some structure to scoped vars.


paulkerschen:
Thanks, @johncrossman, clearer for sure.

PR #387, opened 2017-05-04
COL-861 Derive asset impact score from "active" activities

paulkerschen:
This is an attempt to solve the philosophical problem in https://jira.ets.berkeley.edu/jira/browse/COL-861 without having to change too much logic or modify the database. 

I think the solution is to associate "Impact," for purposes of assets, with "active" activities rather than "passive" activities (we need better terms for this distinction).

"Impact" visualizations for individual users will continue to derive from "passive" activities.


johncrossman:
No points for passive impactfulness?! Some of my best work is done passively. Nonetheless... Thanks, @paulkerschen

PR #388, opened 2017-05-04
COL-860, I'm sad to say, elements must stop dancing in the swimlanes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-860

We rely on `assets.items.length` to show/hide swimlane elements. `isLoading` only controls "No matching assets..." message.

And, for now, comment out "Pinned" filter.


paulkerschen:
Stop your dancing in the swimlanes / Better think of your future

PR #390, opened 2017-05-05
COL-744, asset-library-search gets new filter: by Section

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-744

List of distinct section names can be extracted from the list `users` already available in angular directive. SQL behind search now has `array_position(u.canvas_course_sections, :section)`

### DIFF
--- a/node_modules/col-assets/lib/api.js
+++ b/node_modules/col-assets/lib/api.js
@@ -270,6 +270,7 @@ var addUserToAsset = module.exports.addUserToAsset = function(asset, userId, cal
  * @param  {String}         [filters.keywords]              A string to filter the assets by
  * @param  {Number}         [filters.category]              The id of the category to filter the assets by
  * @param  {Number}         [filters.user]                  The id of the user who created the assets
+ * @param  {String}         [filters.section]               The name of section (i.e., subset of users) to filter assets by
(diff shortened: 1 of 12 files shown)


johncrossman:
image

node_modules/col-assets/lib/api.js
```diff
@@ -367,6 +371,10 @@ var getAssets = module.exports.getAssets = function(ctx, filters, sort, limit, o
     sqlQuery += ' AND au.user_id = :user';
   }
 
+  if (filters.section) {
+    sqlQuery += ' AND (array_position(u.canvas_course_sections, :section) > 0)';
+  }
+
```
johncrossman: Postgres doc: https://www.postgresql.org/docs/9.5/static/functions-array.html

paulkerschen:
Thanks, @johncrossman, code looks good to me. I hate to bring this up since our test framework is so difficult to work with, but what do you think about adding some coverage for the back-end component in `col-assets/tests/test-assets.js`? We have a utility method set up to verify search results, so the grunt work would involve creating test users and associating them with different sections.

johncrossman:
@paulkerschen I accept that challenge. It'll be another climb up the learning curve.

johncrossman:
I'm closing this PR in favor of upcoming PR w/ test coverage. Please stand by.

PR #391, opened 2017-05-05
COL-801, Get back to Impact Studio ... Get back Jojo

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-801


paulkerschen:
But I can't see myself moving back to Tucson now, no matter what Sir Paul says.

PR #393, opened 2017-05-09
COL-868 Fix Sequelize activity-comment association

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-868

Frameworks, frameworks.


johncrossman:
Gee willikers, frameworks are the best. Thx, @paulkerschen

PR #395, opened 2017-05-09
COL-744, asset-library-search gets new filter: by Section

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-744

List of distinct section names can be extracted from the list `users` already available in angular directive. SQL behind search now has `array_position(u.canvas_course_sections, :section)`

image

Initial PR #390 was closed due to no test coverage.

PR #396, opened 2017-05-10
COL-744, whiteboard > add asset > use existing > search by section

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-744

Follow up to PR #395 

Supports search by section when 'Use Existing', as below.

image

### DIFF
--- a/public/app/assetlibrary/search/search.html
+++ b/public/app/assetlibrary/search/search.html
@@ -69,7 +69,7 @@
               <option value="" selected>Uploader</option>
             </select>
           </div>
-          <div class="col-sm-4" data-ng-if="sections.length">
+          <div class="col-sm-4" data-ng-show="sections.length">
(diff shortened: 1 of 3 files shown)


public/app/assetlibrary/search/search.html
```diff
@@ -69,7 +69,7 @@
               <option value="" selected>Uploader</option>
             </select>
           </div>
-          <div class="col-sm-4" data-ng-if="sections.length">
+          <div class="col-sm-4" data-ng-show="sections.length">
```
johncrossman: Note: Use of `data-ng-if` above frequently setting non-empty 'section' value to empty. The `ng-if` was removing `ng-model` from scope prior to search query. The change above fixes the prob and does not change UX.

paulkerschen:
Everything we need and nothin we don't, thanks @johncrossman.

PR #398, opened 2017-05-11
COL-831 Allow impact scores to be recalculated from stored activities

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-831


johncrossman:
Recalculator the impactulator is spectacular. Thx, @paulkerschen

PR #400, opened 2017-05-11
COL-800, new filters for getAssets (hasImpact, hasLikes, etc.) benefit swimlanes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-800

For example, when `filters.hasImpact` is:
* `true` then exclude assets with zero impact
* `false` then include only assets with zero impact
* `null` then do nothing

These filters are primarily for asset exclusion in swimlanes but coded for general purpose.


paulkerschen:
@johncrossman Does this apply to the Dashboard page only, or will it also affect ordering by "most comments," "most likes," etc. in the Asset Library? I think QA dept recommended both.

johncrossman:
You are right, @paulkerschen I was so busy swimming in the shiny Impact Studio that I forgot about the dusty ol' Asset Library.  Amendment coming...

paulkerschen:
Benefits to all! Impact Bruins! Thanks @johncrossman

PR #401, opened 2017-05-11
COL-863 Use app timezone for EI CSV export

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-863


johncrossman:
Timezone matters. Thx, @paulkerschen

PR #405, opened 2017-05-12
COL-880, click 'Most Impactful' > if zero results then proper message and retain filters

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-880

Sample use case:
image

And then...
image

Suggestions on improved wording are welcome.


paulkerschen:
Thanks for kicking the uninvited guests out of the pool, @johncrossman.

PR #406, opened 2017-05-12
COL-831 Make course argument optional in impact recalculation script

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-831

This enhancement allows recalculation of impactulation for all courses in one go, for example after a database refresh.

Also remove the longstanding annoyance of Node scripts scheduling email notifications on startup. This now happens only when the actual app is initialized.

### DIFF
--- a/node_modules/col-activities/lib/api.js
+++ b/node_modules/col-activities/lib/api.js
@@ -66,8 +66,6 @@ var scheduleDaily = function() {
   setTimeout(DailyNotifications.collect, diff, scheduleDaily);
 };
 
-scheduleDaily();
-
 /**
  * Schedule the collection of weekly email notifications
  *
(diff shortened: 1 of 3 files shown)


node_modules/col-activities/lib/api.js
```diff
@@ -863,60 +868,62 @@ var setUserPoints = function(points, userId, callback) {
...
-  // Get the activity configuration for the course
-  getActivityTypeConfiguration(course.id, function(err, configuration) {
+  // Since impact scores can't be overriden, the default activities configuration suffices. Index by type
+  // for quick lookup.
+  var activityConfigurationByType = _.keyBy(ActivitiesDefaults, 'type');
+
+  // Get all asset-associated activities for the course[s]
+  var options = {
+    'where': {
+      'asset_id': {'$ne': null},
```
johncrossman: @paulkerschen unneeded comma above?

johncrossman:
If/when Travis happy then I'm happy. Thx, @paulkerschen

PR #408, opened 2017-05-13
COL-884, gulp eslint on node_modules/col-* plus fixes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-884

Notes:
* [`.eslintignore` documentation](http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories)
* We leverage a double-negative: Ignore default whereby `node_modules` is ignored.
* Too many violations in `col-activities`, etc. to include fixes in this PR.
* Obsolete scripts are ignored
* These are strictly style fixes.

### DIFF
--- a/scripts/update_lti.js
+++ b/scripts/update_lti.js
@@ -56,13 +56,14 @@ var courseCount = null;
 
 /**
  * Connect to the database and kick off.
+ * @return {void}
  */
 var init = function() {
   require('col-core/lib/globals');
 
-  DB.init(function(err) {
-    if (err) {
-      return log.error({'err': err}, 'Unable to establish a database connection');
+  DB.init(function(dbErr) {
+    if (dbErr) {
+      return log.error({'err': dbErr}, 'Unable to establish a database connection');
     }
 
     log.info('Connected to the database');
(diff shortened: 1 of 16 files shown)


scripts/update_lti.js
```diff
@@ -106,32 +108,32 @@ var updateToolsForCourse = function(course, index, callback) {
...
-      log.error({'err': err, 'course': course.id}, 'Failed to get external tool configurations for course');
+  CanvasAPI.getExternalToolsForCourse(course, function(apiErr, courseTools) {
+    if (apiErr) {
+      log.error({'err': apiErr, 'course': course.id}, 'Failed to get external tool configurations for course');
       results.error = results.error.concat(toolsToUpdate);
       return callback();
     }
 
-    async.eachOfSeries(toolsToUpdate, function(tool, index, done) {
+    async.eachOfSeries(toolsToUpdate, function(tool, toolIndex, done) {
```
johncrossman: @paulkerschen I was careful with this renaming of `index`. Please give it a close look, too.

paulkerschen: It looks right to me as well, @johncrossman. Only thing I'd suggest is that if we rename here, we also rename `index` to `toolIndex` in the sister function `updateToolsForAccount` below; just for consistency.

johncrossman: Can't argue with consistency, @paulkerschen. Amendment is in.

paulkerschen:
`* @return {Lint}`!

PR #409, opened 2017-05-15
COL-886, Asset Library > Adv Search > sort by Most Impactful

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-886

UX might change.  For now:

image


paulkerschen:
Value="impact", impact="value". I'm convinced, @johncrossman.

PR #410, opened 2017-05-15
COL-889 COL-891 Add activity breakdown counts for user and course to user activities feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-889
https://jira.ets.berkeley.edu/jira/browse/COL-891


johncrossman:
The summary is a mouthful but the code is concise. Thanks, @paulkerschen

PR #412, opened 2017-05-16
COL-889 Avoid crash on activities with no defined impact category

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-889

Fixes bug introduced in #410.


johncrossman:
Bang, boom, crash! Unsafe impact averted. Thx, @paulkerschen

PR #413, opened 2017-05-16
COL-900, Impact Studio > Alter db: add personal_bio column in 'users' table

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-900

Test coverage will come with changes to UsersAPI, next PR. Are we happy with this column naming?


paulkerschen:
@johncrossman, my first choice would have been `the_deets` but `personal_bio` works too.

PR #414, opened 2017-05-17
COL-901, read/write personal_bio via UsersAPI

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-901

`/users/me/personal_bio` added to UsersAPI and corresponding tests.


paulkerschen:
Thanks, @johncrossman, looks like all the magic words are in the right places.

PR #415, opened 2017-05-17
COL-892 COL-893 Front end for activities breakdown

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-892
https://jira.ets.berkeley.edu/jira/browse/COL-893

contributions_bar


johncrossman:
The Impact Studio! It slices, it dices... And that's not all! Well done, @paulkerschen

PR #416, opened 2017-05-17
COL-905 Let everyone see everyone else's activity timeline

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-905


johncrossman:
Set the data free! Thx, @paulkerschen

PR #420, opened 2017-05-18
COL-894, there can be no Impact without a Studio

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-886

Silly rabbit, "impact" is for the special people.


johncrossman:
Do not merge until Travis is green and user testing is complete. cc: @paulkerschen

paulkerschen:
How true, @johncrossman.

PR #423, opened 2017-05-19
COL-922, char limit on personal_bio gets a bump

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-922

The only thing better than 140 chars is 255 chars.


paulkerschen:
Thanks for making us greater than Twitter, @johncrossman.

PR #424, opened 2017-05-19
COL-921, user description gets linky with it. #hashtags, too

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-921

image


paulkerschen:
#LivingLinky

PR #427, opened 2017-05-19
COL-923 Grab, grab, grab went the cursor

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-923

Browser permitting, the mouse cursor on activity timelines should not POINT with ONE finger but GRAB with MANY fingers.


johncrossman:
The grabbing hands grab all they can.  Thx, @paulkerschen

PR #429, opened 2017-05-21
COL-924, tool-href support in hashtagFilter; toolHref logic to shared utilService

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-924

Notes:
* Delicate refactor of toolHref logic such that the new `toolHrefHashtagFilter` can build proper links in, for example, personal descriptions.
* Because dots and dashes are allowed in hashtags, we must tweak the regex which extracts `_id` from toolHref URL.
* Hashtag followed by period (e.g., "Ready for the #world.") leads to search on keyword='world.' but user prolly wanted keyword='world'   Both hashtag-related filters are fixed. 

Bonus: `data-course` is not needed in the markup of directive.

### DIFF
--- a/public/app/app.bootstrap.js
+++ b/public/app/app.bootstrap.js
@@ -61,7 +61,7 @@
     }.bind({}))[0];
 
     // '_id' represents an asset, user or view requested via link action.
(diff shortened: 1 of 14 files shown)


paulkerschen:
One fewer explicit directive attribute means the humans might yet win this one. Thanks @johncrossman.

PR #430, opened 2017-05-22
COL-927 Distinguish courses with same Canvas id in different instances

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-927


johncrossman:
Sad to see the melting pot of tenants go. Thx, @paulkerschen

PR #434, opened 2017-05-23
COL-913, db scripts introduce 'pin' activities_type and pinned_user_assets table

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-913

Prior to code changes, I want to confirm proper db schema.

### DIFF
--- a/scripts/20170523-col913/create_table_pinned_user_assets.sql
+++ b/scripts/20170523-col913/create_table_pinned_user_assets.sql
@@ -1,0 +1,18 @@
+-- Linking table records assets pinned by users
+
+CREATE TABLE pinned_user_assets (
+  asset_id integer NOT NULL REFERENCES assets (id),
+  user_id integer NOT NULL REFERENCES users (id),
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL,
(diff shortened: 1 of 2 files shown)


scripts/20170523-col913/create_table_pinned_user_assets.sql
```diff
@@ -0,0 +1,17 @@
+-- Linking table records assets pinned by users
+
+CREATE TABLE pinned_user_assets (
+  asset_id integer NOT NULL REFERENCES assets (id),
+  user_id integer NOT NULL REFERENCES users (id),
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL
```
paulkerschen: @johncrossman It looks like our other many-to-many join tables (`asset_users`, `asset_categories`, `whiteboard_elements`) also include an `updated_at` timestamp. I assume that's a Sequelize default; any reason not to include it here?

johncrossman: @paulkerschen Having `updated_at` in a linking table did not make sense to me – thus I left it out. That said, consistency and the conventions of Sequelize matter. I'll bring it back.

paulkerschen:
Thanks for pinning the schema on the elephant, @johncrossman.

PR #435, opened 2017-05-23
COL-906 Periodically recalculate asset trending scores from recent activity

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-906

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -49,6 +49,10 @@
     "apiKey": "some shared secret"
   },
   "timezone": "America/Los_Angeles",
+  "trendingScores": {
+    "activityWindow": 168,
+    "recalculationInterval": 60
+  },
   "whiteboardThumbnails": {
     "timeout": 300
   }
--- a/node_modules/col-activities/lib/api.js
+++ b/node_modules/col-activities/lib/api.js
@@ -48,7 +48,7 @@ var WeeklyNotifications = require('./notifications/weekly');
  *
  * @api private
  */
-var scheduleDaily = function() {
+var scheduleDailyNotifications = function() {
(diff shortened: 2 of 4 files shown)


config/default.json
```diff
@@ -49,6 +49,10 @@
     "apiKey": "some shared secret"
   },
   "timezone": "America/Los_Angeles",
+  "trendingScores": {
+    "activityWindow": 168,
+    "recalculationInterval": 60
```
paulkerschen: As noted in comments below, `activityWindow` uses hour units and `recalculationInterval` uses minutes. The idea was to make it easy to override in QA.

node_modules/col-activities/lib/api.js
```diff
@@ -1006,6 +1023,92 @@ var setImpactScore = function(score, assetId, callback) {
...
+  if (course) {
+    log.info({'course': course.id}, 'Will update trending scores for a single course');
+    activityOptions.where.course_id = course.id;
+  } else {
+    log.info('Will update trending scores for all courses');      
+  }
+  
+  DB.Activity.findAll(activityOptions).complete(function(err, activities) {
+    if (err) {
+      log.error({'err': err}, 'Failed to get activities for the course');
```
johncrossman: @paulkerschen minor detail: Should error message above be sensitive to 'single course' vs. 'all courses'?

paulkerschen: @johncrossman Yes, thanks; amended to log query options as they may come.

johncrossman:
Nice work, @paulkerschen  Let the trending begin

PR #436, opened 2017-05-23
COL-907 Add trending score filtering/ordering to assets API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-907


johncrossman:
Looks right. Thx, @paulkerschen

PR #438, opened 2017-05-24
COL-888 COL-908 Trending front end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-888
https://jira.ets.berkeley.edu/jira/browse/COL-908


johncrossman:
Looks good, @paulkerschen 

One question: Does `utilService.buildSearchResultsMessage` need to know about such specialized sort by options? I.e., explain to user why results are zero.  It's a question for UX-perts, I suppose.

paulkerschen:
Whether by accident or design, the formatted `utilService.buildSearchResultsMessage` message currently appears only in screenreader alerts.

PR #440, opened 2017-05-24
COL-931 Refactor trending calculation for performance

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-931
- Only query for asset data that actually affects recalculation;
- Don't instantiate an unnecessary number of models;
- Only execute a database update when data has changed.


johncrossman:
When in doubt, raw SQL.  Thx, @paulkerschen

PR #444, opened 2017-05-25
COL-663 Front-end error handling

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-663

Turns out, if you commit to handling one error in the front end, you have to commit to handling lots of errors. No-cookie 401s can be caused by restrictive Safari settings, but also show up in other contexts, such as trying to load SuiteC outside an iframe. We show Safari instructions as designed in COL-663 only if SuiteC is in a Safari iframe; otherwise we show a more generic message.

screen shot 2017-05-25 at 1 32 38 pm

In addition, we may be forced to handle errors other than 401s, which we dispatch with an even more generic message.


public/index.html
```diff
@@ -27,6 +27,7 @@
   <link rel="stylesheet" href="/app/engagementindex/leaderboard/userinfo.css" />
   <link rel="stylesheet" href="/app/engagementindex/points/points.css" />
   <link rel="stylesheet" href="/app/shared/activityTimeline.css" />
+  <link rel="stylesheet" href="/app/shared/apiError.css" />
```
johncrossman: @paulkerschen file name above does not match `app/shared/apierror.css`

public/app/shared/apiErrorController.js
```diff
@@ -0,0 +1,50 @@
...
+ * SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
+ * "AS IS". REGENTS HAS NO OBLIGATION TO PROVIDE MAINTENANCE, SUPPORT, UPDATES,
+ * ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+(function(angular) {
+
+  'use strict';
+
+  angular.module('collabosphere').controller('ApiErrorController', function(deviceDetector, apiError, $scope) {
```
johncrossman: @paulkerschen the ghost of Vuerings wants you to alphabetize! (Not a merge blocker)

public/app/shared/apiErrorController.js
```diff
@@ -0,0 +1,50 @@
...
+    // Instructions specific to Safari LTI tools should only be displayed if 1) the browser is Safari and 2) SuiteC
+    // is being run in an iframe.
+    $scope.safariLti = ((window.top !== window.self) && deviceDetector.browser === 'safari');
```
johncrossman: @paulkerschen inner or outer parens, but not both?

public/index.html
paulkerschen: Thanks, @johncrossman, I was about due to revisit https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git ... amended.

public/app/shared/apiErrorController.js
paulkerschen: Sorry, V!

public/app/shared/apiError.html
```diff
@@ -0,0 +1,58 @@
...
+          SuiteC needs to set its cookies in the Safari web browser before the SuiteC tools can be used. (Note: other browsers do not require these permissions.)
+        </div>
+        <div data-ng-if="step === 2">
+          Under the Safari browser menu bar, select the <strong>"Preferences..."</strong> option.
+        </div>
+        <div data-ng-if="step === 3">
+          Select <strong>"Privacy"</strong> and choose <strong>"Always Allow"</strong> in the <strong>"Cookies and website data"</strong> section.
+        </div>
+        <div data-ng-if="step === 4">
+          You're all set to start exploring the Asset Library, Engagement Index, and Whiteboards. Optional: return your Privacy preferences back to their original settings after using the SuiteC tools.
```
johncrossman: @paulkerschen Why do we not say, "...all set to start impacting the..."  Sad!

johncrossman:
Our first built-in tutorial! I'm sad only because we don't have a Donald Trump whiteboard screenshot.  Thx, @paulkerschen

PR #445, opened 2017-05-26
COL-935, AssetsAPI gets pin and unpin; getAssets query gets filter and sort per pins

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-935

Does not include scoring yet. The `getAssets` query has unorthodox sort-by-pin_count implementation – using strictly SQL would cause a big refactor and risk (i.e., GROUP BY per SELECT criteria).

### DIFF
--- a/node_modules/col-assets/lib/api.js
+++ b/node_modules/col-assets/lib/api.js
@@ -148,6 +148,10 @@ var getAsset = module.exports.getAsset = function(ctx, id, opts, callback) {
           'model': DB.User,
           'attributes': UserConstants.BASIC_USER_FIELDS
         }]
+      },
+      {
+        'model': DB.Pin,
+        'as': 'pins'
       }
     ]
   };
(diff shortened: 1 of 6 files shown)


node_modules/col-assets/lib/api.js
```diff
@@ -395,35 +403,53 @@ var getAssets = module.exports.getAssets = function(ctx, filters, sort, limit, o
...
+    DISTINCT ON (a.id, a.likes, a.views, a.comment_count, a.impact_score, a.trending_score)
+    a.*,
+    act.type AS activity_type,
+    (select count(*) from pinned_user_assets p where p.asset_id = a.id)::int AS pin_count
+  `;
+  resultsSqlQuery += sqlQuery;
+
+  // A nasty SQL 'GROUP BY' clause is needed to sort by `pin_count`.
+  // Instead, sort by pin_count after query execution.
+  var sortByPinned = (sort === 'pins');
```
paulkerschen: @johncrossman I think the problem with sorting after query execution is that it won't play nicely with the LIMIT and OFFSET that we use to retrieve assets in groups of 10. I'd expect this to sort assets by pin_count within each group of 10, but the groups of 10 themselves would come back without explicit ordering--which might even do something odd like return the oldest groups first. Am I missing something? If not, the choice might be between a) nasty SQL and b) not offering pin_count as a sort option. I think b) is honorable.

johncrossman: Thanks, @paulkerschen  I'll revisit tomorrow and decide on an honorable choice.

johncrossman:
@paulkerschen Sort by pin_count is gone. Based on current reqts, I don't think it's necessary.

paulkerschen:
Pinxquisite, @johncrossman.

PR #447, opened 2017-05-27
COL-915, UI changes (icons, etc.) in Asset Library per pinning

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-915

UX: https://jira.ets.berkeley.edu/jira/browse/COL-808

List and detail view in Asset Library get basic pinning function.

### DIFF
--- a/node_modules/col-assets/lib/api.js
+++ b/node_modules/col-assets/lib/api.js
@@ -411,6 +411,9 @@ var getAssets = module.exports.getAssets = function(ctx, filters, sort, limit, o
   var hasPins = CollabosphereUtil.getBooleanParam(filters.hasPins, null);
   if (hasPins !== null) {
     sqlQuery += hasPins ? ' AND p.asset_id IS NOT NULL' : ' AND p.asset_id IS NULL';
+    if (hasPins && filters.user) {
+      sqlQuery += ' AND p.user_id = :user';
+    }
   }
 
   var hasTrending = CollabosphereUtil.getBooleanParam(filters.hasTrending, null);
@@ -427,11 +430,9 @@ var getAssets = module.exports.getAssets = function(ctx, filters, sort, limit, o
   var resultsSqlQuery = `SELECT
     DISTINCT ON (a.id, a.likes, a.views, a.comment_count, a.impact_score, a.trending_score)
     a.*,
-    act.type AS activity_type,
-    (select count(*) from pinned_user_assets p where p.asset_id = a.id)::int AS pin_count
+    act.type AS activity_type
   `;
   resultsSqlQuery += sqlQuery;
-
 
   if (sort === 'recent') {
     resultsSqlQuery += ' ORDER BY a.id DESC';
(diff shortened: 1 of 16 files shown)


public/app/assetlibrary/assetLibraryService.js
```diff
@@ -74,24 +74,40 @@
...
+          }
+        });
+      }
+    };
+
+    considerRedirect();
+
+    return {
+      'putPinState': putPinState
+    };
```
johncrossman: As you might suspect, the above is more readable with `?w=1`

node_modules/col-assets/lib/api.js
```diff
@@ -513,7 +517,25 @@ var getAssets = module.exports.getAssets = function(ctx, filters, sort, limit, o
           });
         });
 
-        return callback(null, data);
+        var options = {
+          'where': {
+            'asset_id': assetIds
+          }
+        };
+        DB.Pin.findAll(options).complete(function(err, pins) {
```
paulkerschen: @johncrossman I might be off on this, but I think this additional query might not be necessary since we already have an asset-to-pins join specified in the big old raw query for assets. See here:
https://github.com/sequelize/sequelize/issues/1830
(toward the bottom of the thread, where answers start coming), for the right magic options to have Sequelize parse a raw query into models with associations.

public/app/shared/utilService.js
```diff
@@ -456,14 +456,32 @@
      * @param  {String}     searchOptions        Type of sort selected in search/swimlane
      * @return {void}
      */
-    var addAssetInclusionFilter = function(searchOptions) {
-      if (searchOptions && searchOptions.sort && searchOptions.sort !== 'recent') {
-        searchOptions['has' + _.startCase(searchOptions.sort)] = true;
+    var narrowSearchPerSort = function(searchOptions) {
```
paulkerschen: That's a good name.

node_modules/col-assets/lib/api.js
johncrossman: Good tip, @paulkerschen   I'll dig in tomorrow morning.

johncrossman: @paulkerschen Okay, no luck here. I've tested with permutations similar to suggestion 
 in sequelize PR 1830 above. Few observations:
* This block has a similar `DB.User.findAll` nested under the raw SQL. Do you think it's also unnecessary?
* I notice that other definitions for `Db.Asset` model (e.g., Category) do not influence `getAssets` results.  Whereas, those definitions do matter in the more textbook `getAsset` case.
* I just removed the following from raw SQL in question. (pin_count is not needed, after all.) Thus a small performance improvement.
```
(select count(*) from pinned_user_assets p where p.asset_id = a.id)::int AS pin_count
```

I suggest we create a new Jira to measure and improve performance of `getAssets`. In the process, we might eliminate the nested `DB.User.findAll`. Thoughts?

paulkerschen: @johncrossman Thanks for trying things out; and agreed, let's create a new Jira for any further work in that direction.

Per comments in code, I think the nested `DB.User.findAll` is necessary because it retrieves all associated users even when we're filtering by a given user: so it may return more information than was present in the original query.

johncrossman: @paulkerschen da Jira: https://jira.ets.berkeley.edu/jira/browse/COL-952

PR #450, opened 2017-05-31
COL-949 Badges: back end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-949

All this to get a boolean in the asset feed. `?w=1` review recommended as ever.

Due to the required omission of instructor assets, the query for a 90th-percentile cutoff score is somewhat expensive. If there's a noticeable impact in production we may need to store and recalculate the cutoff per course.


johncrossman:
image

If it's good enough for Cream...  Thanks, @paulkerschen

PR #451, opened 2017-05-31
NOJIRA Sample Mocha approach

raydavis:
For review, not necessarily for merging.

This is by way of getting my first day's homework checked before I move to the LRS. @johncrossman , @paulkerschen , @pfarestveit , @sandeepmjay , would you mind taking a look? Thanks!

### DIFF
--- a/node_modules/col-assets/tests/test-bookmarklet.js
+++ b/node_modules/col-assets/tests/test-bookmarklet.js
@@ -24,6 +24,7 @@
 */
 
 var assert = require('assert');
+var async = require('async');
 var randomstring = require('randomstring');
 
 var AssetsTestUtil = require('./util');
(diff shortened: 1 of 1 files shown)


node_modules/col-assets/tests/test-bookmarklet.js
```diff
@@ -33,155 +34,97 @@ var UsersAPI = require('col-users');
...
+      callback();
+    });
+  });
+  it('creates links with optional metadata', function(callback) {
+    var opts = {
+      'description': 'University of California, Berkeley homepage',
+      'source': 'http://www.universityofcalifornia.edu/uc-system'
+    };
+    AssetsTestUtil.assertCreateLinkBookmarklet(bookmarkletClient, testCourse, testMe.id, testMe.bookmarklet_token, 'UC Berkeley', 'http://www.berkeley.edu/', opts, function(asset) {
+      callback();
```
paulkerschen: @raydavis While it doesn't make a lick of difference in these test cases, we've all had drilled into our heads the convention of an explicit `return` before callbacks whenever possible, just to reduce the risk of hard-to-track bugs where a function keeps going after the callback is called.

paulkerschen:
This is definitely easier on the eyes than the code it replaces, and I'd be happy to see it make its way into the codebase.

node_modules/col-assets/tests/test-bookmarklet.js
raydavis: Thank you, @paulkerschen . I was wondering about that convention & the explanation makes sense. I'll amend, since you're pro-merge.

While we're on the subject of conventions, I note that Mocha's docs always name their async "I-am-all-done-with-my-test" function argument "done". As a newcomer, that made sense to me as a way to distinguish it from customer-authored callback functions. Do you know why PDG uses the more generic name?

paulkerschen: @raydavis, I don't know why the `callback` naming convention was adopted as far as Mocha tests in particular. As a general rule, the convention in the codebase is that `callback` is used for passed-in callback functions, while `done` is used within functions if a different, non-conflicting callback needs to be defined. Usually this happens because of control flow with `async`, for example here: https://github.com/ets-berkeley-edu/suitec/blob/master/node_modules/col-activities/tests/util.js#L336

PR #454, opened 2017-06-02
COL-950 Badges: front end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-950

The blue is matched to Canvas navigation tabs. See screenshot:

top_asset


johncrossman:
image

PR #455, opened 2017-06-02
COL-957, pin/unpin from asset-icon-bar; listen for assetPinEventByMe events

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-957

Click pin icon below asset preview to toggle status of myPin. Doing so fires `assetPinEventByMe` events which, in some cases, cause search results (or profile page swimlane) to reload appropriately.

**Bonus:** "Uploader" is now "User" per https://jira.ets.berkeley.edu/jira/browse/COL-953

### DIFF
--- a/public/app/assetlibrary/list/list.html
+++ b/public/app/assetlibrary/list/list.html
@@ -60,8 +60,8 @@
       </li>
       <!-- LIST ITEM -->
       <li class="assetlibrary-list-item list-inline col-xs-6 col-sm-4 col-md-3" data-ng-repeat="asset in assets | unique:'id'">
-        <a ui-sref="assetlibrarylist.item({assetId: asset.id})">
-          <div class="col-list-item-container">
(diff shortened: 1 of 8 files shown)


public/app/assetlibrary/list/list.html
```diff
@@ -83,9 +83,9 @@
                 </div>
               </div>
             </div>
-            <div class="text-right col-list-item-actions" iconbar asset="asset" show-pin="true" viewonly="true"></div>
-          </div>
-        </a>
+          </a>
+          <div class="text-right col-list-item-actions" iconbar asset="asset" show-pin="true" viewonly="true"></div>
+        </div>
```
johncrossman: `?w=1` for a simpler diff above.

paulkerschen:
Thanks @johncrossman; in pinning assets it's important to be a good listener.

PR #458, opened 2017-06-02
COL-969, profile page and asset-iconbar get ids on buttons and hrefs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-969

A chicken in every pot and an `id` on every button.


paulkerschen:
Robot friendly!

PR #459, opened 2017-06-02
COL-929 Omit activities on deleted assets from activity feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-929

The trick: activities on deleted assets should be excluded, but activities with no asset should still show up. We first retrieve, then suppress.


johncrossman:
Yes we want no deleted. Thx, @paulkerschen

PR #460, opened 2017-06-02
COL-969, tool-href directive cannot have dynamic id, data-ng-attr-id or otherwise

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-969

Another Angular lesson learned. Fortunately, QA is more concerned with having ids on _buttons_.

### DIFF
--- a/public/app/dashboard/profile.html
+++ b/public/app/dashboard/profile.html
@@ -226,8 +226,7 @@
         <li class="assetlibrary-list-item list-inline col-xs-6 col-sm-4 col-md-3"
             data-ng-repeat="asset in user.assets.results">
           <div class="col-list-item-container">
-            <a id="user-asset-{{asset.id}}"
-              tool-href
+            <a tool-href
               data-tool="assetlibrary"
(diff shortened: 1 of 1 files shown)


public/app/dashboard/profile.html
```diff
@@ -226,8 +226,7 @@ <h4 class="profile-activity-breakdown-header">Community Impacts</h4>
         <li class="assetlibrary-list-item list-inline col-xs-6 col-sm-4 col-md-3"
             data-ng-repeat="asset in user.assets.results">
           <div class="col-list-item-container">
-            <a id="user-asset-{{asset.id}}"
-              tool-href
+            <a tool-href
```
johncrossman: Fyi, `data-ng-attr-id` would not work here. The way to put dynamic ids on these directive-driven tags is via directive itself. I.e., in `toolHrefDirective`.  Again, more trouble than it's worth.

paulkerschen:
Thanks for the object lesson, @johncrossman - sneaky!

PR #462, opened 2017-06-02
COL-933 Adjust to saner EventDrops dimensions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-933

A bit of compensation for changes in https://github.com/ets-berkeley-edu/EventDrops/pull/8, which should be merged first.


johncrossman:
Sanity checked. Thx, @paulkerschen

PR #466, opened 2017-06-05
COL-977, deal with event 'assetPinEventByMe' regardless of sortBy

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-977

The "400 error" described in Jira was due to pin=true attempt on an already-pinned asset. If user pins asset on his/her profile we must _always_ refresh.


paulkerschen:
Sneaky. Thanks @johncrossman

PR #468, opened 2017-06-05
COL-947 Do not record asset library searches in analytics when loading Impact Studio

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-947

What is an Asset Library search? Chocolate.

What is an Impact Studio profile? Peanut butter.

You know what to do.


johncrossman:
This looks good, @paulkerschen  Do we want the same `track=false` in whiteboardsReuseController?

PR #469, opened 2017-06-05
COL-983, tool-href don't play well with ids, part 2

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-983

I scrutinized all other `tool-href` usages. We _should_ be back in business.

Use `w=1`


paulkerschen:
Left your ID at home, did you, son?

PR #470, opened 2017-06-05
COL-947 More Impact Studio activity tracking

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-947


johncrossman:
I can see no wrong... Thx, @paulkerschen

PR #471, opened 2017-06-05
NOJIRA Don't make assumptions about other people's hobbies

paulkerschen:


johncrossman:
Some are unhappy with this change, @paulkerschen 

image

PR #474, opened 2017-06-06
COL-988, getAssets: 'order by pins' is, to be more specific, ordered by pin.created_at desc

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-988

Because it just makes sense.


paulkerschen:
Thanks for proper pinning priority, @johncrossman.

PR #476, opened 2017-06-06
COL-955 Asset detail buttons should not take up two parking spots in the Bootstrap grid

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-955

"Pin" just isn't a very long word.


johncrossman:
Sadly, "Pin" has short-word-syndrome. ... Thx, @paulkerschen

PR #478, opened 2017-06-06
COL-986 Wrap, don't truncate, activity breakdown labels

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-986

Cascade, sweet stylesheets, till I end my song.


johncrossman:
All that scratchin' is making me itch. ... Break it on down, @paulkerschen

PR #479, opened 2017-06-06
COL-989, activityTimelineDirective will refresh when model object is updated

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-989

Refresh directive when `data-activity-timeline` is modified.

I recommend `w=1`.


paulkerschen:
Whole lotta sense, @johncrossman, thanks.

PR #481, opened 2017-06-06
COL-940 Disable metaballs on Safari

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-940

"The staff was friendly but the rendering was soggy... would not go back"


johncrossman:
image

PR #483, opened 2017-06-07
COL-988, sort order :user_id must match SELECT clause :user_id

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-988

Pin sort by user when user is not me gives pins sorted as if you were me. Oh, never mind.


paulkerschen:
notme

PR #485, opened 2017-06-07
COL-986 Fix alignment on activity breakdown labels

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-986

Fun fact! If the text all fits on one line, fancy flexbox `justify-content: flex-end;` controls alignment. If the text wraps, boring old `text-align: right;` must be used.


johncrossman:
Sometimes, you gotta go old school. Thx, @paulkerschen

PR #486, opened 2017-06-07
COL-992 Update timeline start/end on activity refresh

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-992


johncrossman:
Ah, yes. Thx, @paulkerschen

PR #487, opened 2017-06-07
COL-992 Fix ordering in activity feeds

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-992

Sequelize must be asked extra nice when ordering on associations.


johncrossman:
Pretty please, Sequelize!

PR #488, opened 2017-06-08
COL-998 Impacts bar chart should be colored Impact Rouge

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-998

impacts

It matches the eventdrops, don't you know.


johncrossman:
Bloody brilliant, @paulkerschen

PR #489, opened 2017-06-08
COL-990, search-for-people if you're an admin who is not part of the roster, etc.

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-990

Course with...
* less than 2 users gets no search browse
* two users gets UX below
* three or more users gets standard experience

I recommend `?w=1`

---
image


paulkerschen:
There's a frightening story in here about a course with only two users forced to endlessly impact each other. Thanks @johncrossman.

PR #493, opened 2017-06-09
COL-972, User with 1+ pins and zero uploaded assets needs filters in 'Assets' swimlane

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-972

Expected UI of 'My Assets':

----
image

Expected UI of 'Assets' (some other user):

----
image


paulkerschen:
They who pin and upload not are also impactful in their way. Thanks @johncrossman.

PR #494, opened 2017-06-09
COL-1004, test CodeBuild env var: CODEBUILD_RESOLVED_SOURCE_VERSION

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1004

CodeBuild succeeded but post_build phase complained:
```
Not a git repository (or any parent up to mount point /codebuild/output)
```

[AWS docs](http://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref.html#build-env-ref-env-vars) suggest `CODEBUILD_RESOLVED_SOURCE_VERSION`. This PR checks that value and moves `generate-buildspec-report.sh` to build phase.

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -15,11 +15,11 @@ phases:
       - ./node_modules/.bin/gulp eslint
   build:
     commands:
-      - echo "build phase"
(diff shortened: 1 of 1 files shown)


paulkerschen:
Sure!

PR #496, opened 2017-06-10
COL-987, swim lanes must settle down when you toggle between empty and non-empty lanes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-987

Put fixed min-height on swim lanes. In addition to herky-jerky UI, the "No matching assets" alert flickered when toggling between lanes.  This was fixed by removing the use of `isLoading`. 

The most dramatic case of empty real estate is below. This is only seen when user has zero uploaded assets. The empty space is a call to action!

image

----

image

### DIFF
--- a/public/app/dashboard/profile.css
+++ b/public/app/dashboard/profile.css
@@ -237,7 +237,11 @@
 }
 
 .profile-assets-container {
-  margin-bottom: 30px;
+  min-height: 360px;
+}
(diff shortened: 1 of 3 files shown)


paulkerschen:
Looks good @johncrossman. How can the swimlanes think of dancing at a time like this?

PR #502, opened 2017-06-13
COL-1010, use of UsersAPI.getUsers to get user data is problematic

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1010

The `me` object (set in col-core module with `UsersAPI.getUser`) has all data we need. However, `otherUsers` list comes from `UsersAPI.getUsers` which is less rich. Two fixes here:
1. Front-end profileController wipes the slate fully clean when browsing to next user.
2. The `/users/id/:id` endpoint now relies on the richer result-set of `UsersAPI.getUser`.


paulkerschen:
Thanks, @johncrossman, cleaner for sure.

PR #503, opened 2017-06-14
COL-1012 Use elbv2 settings for load balancer SSL config

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1012

The new settings are, thankfully, rather simpler.

### DIFF
--- a/.ebextensions/01_create_apache_conf.config
+++ b/.ebextensions/01_create_apache_conf.config
@@ -4,7 +4,7 @@
 files:
 
   # The default Elastic Beanstalk .conf file for NodeJS containers must be overwritten. We create a
-  # replacement file here, then overwrite the default (02_overwrite_apache_conf.config) and restart Apache 
+  # replacement file here, then overwrite the default (02_overwrite_apache_conf.config) and restart Apache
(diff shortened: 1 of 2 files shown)


.ebextensions/01_create_apache_conf.config
```diff
@@ -126,7 +126,6 @@ files:
         SSLSessionTickets     Off
 
         Header always set Strict-Transport-Security "max-age=63072000; includeSubdomains; preload"
-        Header always set X-Frame-Options DENY
```
paulkerschen: Note that this had to be taken out since it prevented using the tools within an iframe. The fact that it's been hanging out in our ssl.conf for months implies that up until now the entire ssl.conf was being ignored.

sandeepmjay: @paulkerschen : Not having the X-Frame_Options resulted in few errors on the LRS side of things. I used this as a way to get around the config.

"Header always unset X-Frame-Options"

johncrossman:
Thank you for getting this done, @paulkerschen  You've earned a DevOps "Badge of Impact"

PR #504, opened 2017-06-14
COL-802, new tool-href args (scroll, referringScroll) for proper focus when Return to Dashboard

johncrossman:

Worth noting:
*  profileController must scroll-to-position _after_ all assets loaded, thus the callbacks
* `crossToolRequest` instead `referringTool` allows us to ditch prefix 'requested' on id and scroll
* use `?w=1`

### DIFF
--- a/public/app/app.bootstrap.js
+++ b/public/app/app.bootstrap.js
@@ -49,30 +49,43 @@
   ]);
 
   /**
+   * Extract parameter value from URI using regex pattern
+   *
+   * @param  {String}       uri           Query string of current URL
+   * @param  {Object}       pattern       Regex for identifying parameter value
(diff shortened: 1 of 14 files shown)


public/app/app.bootstrap.js
```diff
@@ -98,24 +111,15 @@
       collabosphere.constant('me', results.me.data);
       collabosphere.constant('config', results.config.data);
       collabosphere.constant('apiError', null);
+      collabosphere.constant('crossToolRequest', params.crossToolRequest || null);
```
johncrossman: A name better than `crossToolRequest`? Suggestions welcome.

paulkerschen:
I like `crossToolRequest`, @johncrossman. Thanks for the extra cleanup.

PR #506, opened 2017-06-14
COL-802, ignore scroll when search/browse other users

johncrossman:

`scroll` is one property we want to forget when browsing other users.

### DIFF
--- a/public/app/dashboard/profileController.js
+++ b/public/app/dashboard/profileController.js
@@ -78,6 +78,8 @@
         analyticsService.track('Search for user profile', {
           'search_for_user': $scope.browse.searchedUserId
         });
+
+        crossToolRequest.scroll = null;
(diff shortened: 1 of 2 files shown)


public/app/dashboard/profileController.js
```diff
@@ -78,6 +78,8 @@
         analyticsService.track('Search for user profile', {
           'search_for_user': $scope.browse.searchedUserId
         });
+
+        crossToolRequest.scroll = null;
         $state.go('userprofile', {'userId': $scope.browse.searchedUserId});
```
johncrossman: We can't use `loadProfileById(...)` in the line above because the `oi-select` (search input) would not be refreshed.

paulkerschen:
Consider the scroll, for wisdom is writ there. Thanks @johncrossman

PR #508, opened 2017-06-15
COL-1015 Update Sendgrid version and API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1015

Per examples at https://github.com/sendgrid/sendgrid-nodejs, code needed to interact with Sendgrid looks rather different these days.


johncrossman:
Looks good. Watching on Dr. T, @paulkerschen

PR #514, opened 2017-06-16
COL-802, 'Return to Dashboard' restores selected filters and previous scroll position

johncrossman:

If we return user to previous scroll position then we must also restore previous filters. This PR repurposes `referringScroll`.  We use `referringState` to serialize page state.


public/app/dashboard/profileController.js
```diff
@@ -396,10 +462,12 @@
     var init = function() {
       // Determine user
       var userId = $stateParams.userId || (crossToolRequest && crossToolRequest.id);
+      var loadPreviousState = $stateParams.loadPreviousState !== 'false';
```
paulkerschen: I have satisfied myself that everything really does come out of `$stateParams` as a string.

paulkerschen:
Looks right to me, @johncrossman. Stateful and stately.

PR #515, opened 2017-06-18
COL-802, fix lazy mistake in backtoreferrer.html

johncrossman:

...and pass back `referringState`, too.

### DIFF
--- a/public/app/assetlibrary/backtoreferrer.html
+++ b/public/app/assetlibrary/backtoreferrer.html
@@ -4,7 +4,8 @@
   <a tool-href
      class="col-back"
      data-tool="dashboard"
+     data-id="crossToolRequest.referringId"
+     data-state="crossToolRequest.referringState"
      data-referring-tool="assetlibrary"
-     data-id="crossToolRequest.id"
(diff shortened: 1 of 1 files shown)


public/app/shared/toolHrefDirective.js
```diff
@@ -48,8 +48,7 @@
 
       'link': function(scope, elem, attrs) {
         scope.$watch('id', function() {
-          // 'id' and 'referring-tool' attributes may specify either a variable name (the evaluated value
-          // of which appears under 'scope'), or a string literal (which appears under 'attrs').
+          // Examples of 'id': asset id, user id, view name as listed in app.states
```
paulkerschen: @johncrossman I originally put in the comment above because it took me a while to figure out the `scope || attrs` pattern where the directive was happy to accept either a variable name or a string literal in the referring HTML attribute. Do you think there's value in keeping it?

johncrossman: You've just described the value in keeping it. Thx, @paulkerschen  Amend is in.

PR #517, opened 2017-06-19
COL-991, more responsive user pagination in profile view

johncrossman:

No wrap. No outline. 

image


paulkerschen:
No muss, no fuss. Looks good @johncrossman.

PR #522, opened 2017-06-19
COL-1036 API endpoint and tests for Canvas conversation messaging

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1036

There's not really a good place for this API in our current structure, but the users module is perhaps the least bad.


johncrossman:
Least bad is the best we can do. Thanks, @paulkerschen

PR #523, opened 2017-06-20
COL-1043, eslint upgraded to 4.0.0

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1043

Fyi, http://eslint.org/docs/user-guide/migrating-to-4.0.0


public/app/engagementindex/leaderboard/leaderboardController.js
```diff
@@ -114,7 +114,12 @@
               'backgroundColor': 'transparent',
               'inverted': true,
               // Ensure that the box plot is displayed horizontally
-              'margin': [0, 20, 0, 20],
+              'margin': [
+                0,
+                20,
+                0,
+                20
+              ],
```
johncrossman: The rules I've put in require multi-line arrays when length > 3. In most cases, it's reasonable and good. In a few cases (as above), it's silly. IMO, benefits outweigh the silliness.

paulkerschen:
I don't know what is meant by `no-useless-escape` or `no-confusing-arrow`, but I love the sound of both. Thanks @johncrossman.

PR #524, opened 2017-06-20
COL-1028 User.looking_for_collaborators: database and feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1028

Since this value is expected to surface across the app, including the Engagement Index, it must be added to `BASIC_USER_FIELDS` and `POINTS_USER_FIELDS`.

Suppressing this value from the feed based on Impact Studio use turns out to be impractical, as it would require too many joins to the `courses` table to check for `dashboard_url`. We're better off enforcing that logic in the front end, which always has course properties available.

### DIFF
--- a/config/schema.sql
+++ b/config/schema.sql
@@ -15,14 +15,14 @@ SET client_min_messages = warning;
 SET row_security = off;
 
 --
--- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: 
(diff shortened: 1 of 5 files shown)


johncrossman:
What if I need a corroborator? Nonetheless, looks good, @paulkerschen

PR #525, opened 2017-06-20
COL-914, col-assets/lib/api.js: helper functions moved to util.js

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-914

`col-assets/lib/api.js` was nearing critical mass.


paulkerschen:
Thanks @johncrossman, nothing to argue with here.

PR #526, opened 2017-06-20
NOJIRA, gulp mocha > use MOCHA_GREP or '--grep' arg

johncrossman:
Use MOCHA_GREP or
```
gulp mocha --grep "A subset of all tests"
```


paulkerschen:
Hurrah, more yargs!

PR #527, opened 2017-06-20
COL-1029 API and front-end factory to toggle User.looking_for_collaborators

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1029


johncrossman:
Looks good, back to front. Thx, @paulkerschen

PR #528, opened 2017-06-21
COL-1030 Looking for collaborators: front-end toggle

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1030

Display on other people's profiles comes next.

screen shot 2017-06-21 at 11 56 50 am
screen shot 2017-06-21 at 11 57 02 am

### DIFF
--- a/public/app/dashboard/lookingForCollaborators.html
+++ b/public/app/dashboard/lookingForCollaborators.html
@@ -1,0 +1,17 @@
+<div class="profile-looking-for-collaborators-wrapper">
+  <div data-ng-if="isMyProfile">
+    <span class="profile-looking-for-collaborators-heading">Looking for collaborators:</a>
+    <label class="profile-looking-for-collaborators-toggle">
+      <input type="checkbox"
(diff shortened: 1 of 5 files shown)


public/app/dashboard/lookingForCollaborators.html
```diff
@@ -0,0 +1,17 @@
+<div class="profile-looking-for-collaborators-wrapper">
+  <div data-ng-if="isMyProfile">
```
johncrossman: @paulkerschen What if user is not a member of course?

johncrossman: ...I suppose s/he can upload assets, etc. so...

paulkerschen: @johncrossman It's a good question that might apply to the profile page more broadly; I think non-member admin users can also edit their personal description, etc?

johncrossman:
Useful in looking for Mr. Goodbar. Thanks, @paulkerschen

PR #529, opened 2017-06-21
COL-1031 Collaboration indicator button on other peeps' profiles

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1031

screen shot 2017-06-21 at 1 15 02 pm


johncrossman:
Looking right to me, @paulkerschen

PR #531, opened 2017-06-21
COL-1032 Add Collaborate column to the Engagement Index

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1032

Per designs, this column appears only in courses with the Impact Studio enabled; is sortable in the same way as other columns; includes a small version of the green button for other users; and duplicates the profile page's toggling functionality for oneself.

screen shot 2017-06-21 at 2 35 31 pm


johncrossman:
Looks good, @paulkerschen  But why am I not at the top of leaderboard?

PR #532, opened 2017-06-22
COL-991, other-users-container and profile-filters: @media screen rule (mobile is default) 

johncrossman:

image

Design for mobile first: https://www.w3schools.com/css/css_rwd_mediaqueries.asp

image


paulkerschen:
The power of flex. Thanks for facilitating impact "on the go," @johncrossman.

PR #537, opened 2017-06-22
COL-991, @media rule min-width: 768px per SuiteC standard

johncrossman:

* 768px as used in assetlibrary/item.css and whiteboards/reuse.css 
* do not hide delimiter of browse-user

### DIFF
--- a/public/app/dashboard/profile.css
+++ b/public/app/dashboard/profile.css
@@ -433,7 +433,7 @@
   white-space: pre-wrap;
 }
 
-@media only screen and (min-width: 500px) {
+@media only screen and (min-width : 768px) {
   .other-users-container {
     flex-direction: row;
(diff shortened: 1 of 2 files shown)


public/app/dashboard/profile.css
```diff
@@ -433,7 +433,7 @@
   white-space: pre-wrap;
 }
 
-@media only screen and (min-width: 500px) {
+@media only screen and (min-width : 768px) {
```
johncrossman: Same rule used on asset detail page: [item.css](https://github.com/ets-berkeley-edu/suitec/blob/master/public/app/assetlibrary/item/item.css#L75)

paulkerschen:
Thanks @johncrossman; we are ever more consistent.

PR #543, opened 2017-06-26
COL-1052 Exclude deleted assets from coursewide activity counts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1052


johncrossman:
Raw like sushi.  Looks good, @paulkerschen Waiting on T

PR #541, opened 2017-06-23
COL-1042 Revised explanatory text for activity visualizations

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1042

Front-end code to make your eyes cross!

screen shot 2017-06-23 at 3 33 38 pm

screen shot 2017-06-23 at 3 34 02 pm

screen shot 2017-06-23 at 3 35 21 pm

PR #544, opened 2017-06-26
COL-1042 reuse, n. Second or further use; an instance of this.

paulkerschen:
Late-breaking emendation to #541.

https://jira.ets.berkeley.edu/jira/browse/COL-1042 (design decision in https://jira.ets.berkeley.edu/jira/browse/COL-866).

1838   Extra Globe 23 Aug. 341/2   The re-use of ‘old wrappers and cut twine’, will save only a few thousands of dollars.
1882   Rep. Precious Metals (U.S. Bureau of Mint) i. 650   The water is pumped up to the stamps again for re-use.
1939   Sun (Baltimore) 21 Feb. 14/4   ‘Reprocessed’..should be applied to wool made into a fibrous state for reuse.
1970   Engin. & Med. 139   The needles..will last for hundreds of reuses.
2008   Computer Weekly 2 Sept. 27/4   Because it is costing £4.75bn to collect the LHC data, it would be profligate not to allow reuse.

### DIFF
--- a/public/app/dashboard/activityTimelineLabelDetails.html
+++ b/public/app/dashboard/activityTimelineLabelDetails.html
@@ -10,7 +10,7 @@
     <div data-ng-switch-when="Creations">
       Includes the activities: add new assets, add assets to whiteboard, export whiteboards, and remix whiteboards.
     </div>
-    <div data-ng-switch-when="Recreates">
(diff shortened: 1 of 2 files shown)


johncrossman:
The re-use of old wrappers and cut twine makes for a beautiful whiteboard.  Waiting on T.

PR #547, opened 2017-06-26
COL-956 Add mobile-specific Safari cookie instructions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-956

This goes beyond the scope of the original ticket, which was just to suppress the desktop instructions on mobile, but considering that SuiteC on iOS is now unusable without making this settings change, some explicit instructions seem called for.


johncrossman:
Above and beyond, @paulkerschen   Waiting on T.

PR #553, opened 2017-06-27
COL-1059, if pinner is co-owner of the asset then his collaborators get neither "pin" nor "get_pin" credit

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1059

Rewarding nepotism was a feature, not a bug.


paulkerschen:
Fair is fair, @johncrossman.

PR #548, opened 2017-06-27
COL-1058 Contributions breakdown: textual revisions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1058

Some changes to mollify the stern eye of QA.
- "Compared to Others" is reverted to the more strictly correct "Compared to Everyone";
- Counts for "Everyone" are converted to averages showing a single decimal place (the rationale being that this makes weird zeroes less likely);
- People other than you are not called "you".

screen shot 2017-06-26 at 9 39 18 pm

screen shot 2017-06-26 at 9 50 41 pm

PR #554, opened 2017-06-27
COL-1058 Round coursewide average activity counts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1058

Follow-up to #548.


johncrossman:
The almost empty glass is actually full.  Thx, @paulkerschen

PR #557, opened 2017-06-27
COL-1061, reset user preferences when loading new user profile

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1061

We expect `defaultUserPreferences` to be immutable. This bug was caused by its corruption.


paulkerschen:
A sneaky one, to be sure. Thanks @johncrossman.

PR #564, opened 2017-06-28
COL-1062 Omit administrator activities from visualizations

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1062


node_modules/col-activities/lib/api.js
```diff
@@ -260,6 +260,13 @@ var getActivitiesForUserId = module.exports.getActivitiesForUserId = function(ct
         return;
       }
 
+      // Omit activities performed by administrators not active in the course.
+      if (activity.actor &&
+          activity.actor.canvas_enrollment_state === 'inactive' &&
+          (_.includes(CollabosphereConstants.ADMIN_ROLES, activity.actor.canvas_course_role))) {
+        return;
+      }
```
johncrossman: @paulkerschen questions on conditional above:
1. does it make sense to have the slightly more broad `(state === inactive || is_admin)`?
1. parens around `_includes()` are unnecessary?

paulkerschen: @johncrossman 
1. I think `(state === inactive || is_admin)` would catch cases like teachers who were originally part of a course site and then left it; which we might not want to screen out in the W4C context for instance.
2. I guess they are!

johncrossman:
Looks good, @paulkerschen  Waiting on Mr. T.

PR #565, opened 2017-06-28
COL-1069, no 'No assets' msg until NOT isLoading

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1069

We only need this check on initial page load.


paulkerschen:
Looks good, @johncrossman, waiting on T.

PR #570, opened 2017-06-29
COL-1073, remove 'npm install' and 'gulp eslint' from buildspec (CodePipeline)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1073

The true test will be CodePipeline.


paulkerschen:
Pipe it!

PR #574, opened 2017-06-29
COL-1076, asset detail gets a new lane: Pinned

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1076

More powder blue dots for asset detail.


paulkerschen:
Pinned, and pinned.

PR #575, opened 2017-06-30
COL-1082 Type check in assetViewRedirect

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1082


johncrossman:
Looks right, @paulkerschen  Waiting on T

PR #579, opened 2017-07-03
COL-1081 Add reciprocal ids to activities db table

paulkerschen:
The solution to https://jira.ets.berkeley.edu/jira/browse/COL-1081 is to create `get_view_asset` activities for every `view_asset` activity that is lacking them. However, this requires an association between reciprocal activities, which doesn't currently exist in the database. This PR creates that association.

Passive activities of a `get_*` type are given a many-to-one association to their corresponding "active" activity. Going forward, this association will be tracked as part of activity creation, which in some cases involves changing the order of those creations.

Adding reciprocal ids to older activities is trickier, because we have to infer the associations based on user ids, timestamps, and other metadata. The SQL included in this PR makes a best effort at such inference. I've tested it in suitec-dev and the results seem sane, although 693 activities of `get_*` form, or 0.9% of the total, couldn't find a reciprocal matching the given criteria and were left blank. Such doubtful cases seem to arise when a flurry of simultaneous activities makes timestamp correlation harder; nonetheless, I think this is an acceptable margin, and leaving the doubtful cases blank makes it easy to repair them manually later, if we so choose.

### DIFF
--- a/node_modules/col-activities/lib/api.js
+++ b/node_modules/col-activities/lib/api.js
@@ -481,7 +481,8 @@ var getActivitiesForAssetId = module.exports.getActivitiesForAssetId = function(
  * @param  {Activity}         callback.activity                   The created activity
  */
 var createActivity = module.exports.createActivity = function(course, user, type, objectId, objectType, metadata, actor, callback) {
-  metadata = metadata || {};
+  // Copy metadata object so that modifications don't affect any other activities that may use the same metadata.
+  metadata = _.extend({}, metadata);
 
   // Parameter validation
   var validationSchema = Joi.object().keys({
@@ -523,6 +524,10 @@ var createActivity = module.exports.createActivity = function(course, user, type
   }
   if (actor) {
     activity.actor_id = actor.id;
+  }
+  if (metadata.reciprocalId) {
+    activity.reciprocal_id = metadata.reciprocalId;
+    delete metadata.reciprocalId;
   }
 
   // TODO: Wrap this in a transaction
@@ -870,6 +875,7 @@ var recalculateImpactScores = module.exports.recalculateImpactScores = function(
  * @param  {Number}         asset               The asset to be pinned or repinned
  * @param  {Function}       callback            Standard callback function
  * @param  {Object}         callback.err        An error that occurred, if any
+ * @param  {Activity}       callback.activity   The created activity
  */
 var createPinActivity = module.exports.createPinActivity = function(ctx, asset, callback) {
   var proceed = !_.find(asset.users, {'id': ctx.user.id});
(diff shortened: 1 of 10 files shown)


node_modules/col-activities/lib/api.js
```diff
@@ -481,7 +481,8 @@ var getActivitiesForAssetId = module.exports.getActivitiesForAssetId = function(
  * @param  {Activity}         callback.activity                   The created activity
  */
 var createActivity = module.exports.createActivity = function(course, user, type, objectId, objectType, metadata, actor, callback) {
-  metadata = metadata || {};
+  // Copy metadata object so that modifications don't affect any other activities that may use the same metadata.
+  metadata = _.extend({}, metadata);
```
johncrossman: @paulkerschen do we need to `cloneDeep` metadata prior to extend?  Similar to PR #558

paulkerschen: @johncrossman Thankfully, that shouldn't be necessary here; the modifications to `metadata` below involve deleting top-level properties, so we don't have to worry about nesting.

node_modules/col-assets/lib/api.js
```diff
@@ -1711,15 +1711,19 @@ var like = module.exports.like = function(ctx, assetId, like, callback) {
...
 
-          ActivitiesAPI.createActivity(ctx.course, ctx.user, type, asset.id, CollabosphereConstants.ACTIVITY.OBJECT_TYPES.ASSET, null, null, function(err) {
+          ActivitiesAPI.createActivity(ctx.course, ctx.user, type, asset.id, CollabosphereConstants.ACTIVITY.OBJECT_TYPES.ASSET, null, null, function(err, likeActivity) {
             if (err) {
               return callback(err);
             }
+            if (!likeActivity) {
+              log.error({'user': ctx.user, 'assetId': asset.id}, 'Failed to retrieve newly created like/dislike activity');
+              return callback({'code': 500, 'msg': 'Failed to retrieve newly created like/dislike activity'});
+            }
```
johncrossman: @paulkerschen Why we must we check for existence of newly created activity? Do we trust `createActivity` less than before?

paulkerschen: @johncrossman Here and elsewhere, we're now fetching the `id` property of the newly created activity, which we weren't doing before. So this code path just amounts to a null check so the process doesn't crash if something truly unexpected happens. It is (I hope) very unlikely that it would ever be triggered.

johncrossman:
A delicate dance indeed. Nice work, @paulkerschen  I'll merge once the discussed SQL amendment is in.

paulkerschen:
Thanks very much for the review, @johncrossman. SQL amendment is in.

johncrossman:
Kaboom!

PR #580, opened 2017-07-05
COL-1081 Insert get_view_asset activities where not present

paulkerschen:
Second and much simpler PR for https://jira.ets.berkeley.edu/jira/browse/COL-1081.

Select every `view_asset` activity not already matched to a `get_view_asset` activity; join to the `asset_users` table to get all asset owners; for each asset owner, insert a matching `get_view_asset` activity.

The query has been tested locally with good results.


johncrossman:
PR has been reviewed with good results. Thx, @paulkerschen

PR #582, opened 2017-07-07
COL-1092, 'Return to Asset Library' now uses advSearchId as seen on Impact Studio

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1092

Rather simple fix by targeting only the 'Return to Asset Library' link. The asset deep-link design does not change.  I added `_.pickBy` logic to `utilService.getAdvancedSearchId` such that undefined  and extraneous args are excluded from queryString.

### DIFF
--- a/public/app/assetlibrary/item/assetLibraryItemController.js
+++ b/public/app/assetlibrary/item/assetLibraryItemController.js
@@ -79,6 +79,14 @@
     utilService.setParentHash({'asset': assetId});
 
     /**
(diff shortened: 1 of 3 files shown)


johncrossman:
@paulkerschen This is one step towards a fix. The concern described in COL-1092 still stands: appending anchor to asset library URL will not reload the page.

johncrossman:
I'll revisit this with a more complete solution but, for now, priority is not high.

PR #584, opened 2017-07-11
COL-951 Remove Sequelize sync; test against db schema in source control

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-951

Our dependency on Sequelize's falsely comforting "sync" behavior is removed. Instead, local and Travis test runs create their database schemas directly from SQL. This require a permissions change to the `suitec` user in test environments (see README). Devs will also have to take care to keep their local databases synced up with the latest migrations, since Sequelize will no longer be helpfully attempting to manage it for us.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@ createuser suitec --no-createdb --no-superuser --no-createrole --pwprompt
 
 createdb suitec --owner=suitec
 createdb suitec_test --owner=suitec
+
+# Give SuiteC schema ownership on the test database so that it can reset the schema during tests
+psql suitec_test -c 'alter schema public owner to suitec;'
 ```
 
 ### Node
--- a/config/schema.sql
+++ b/config/schema.sql
@@ -7,27 +7,11 @@
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
-SET idle_in_transaction_session_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SET check_function_bodies = false;
 SET client_min_messages = warning;
 SET row_security = off;
-
---
--- Name: plpgsql; Type: EXTENSION; Schema: -; Owner:
---
-
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
-
-
---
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -53,6 +53,9 @@ createuser suitec --no-createdb --no-superuser --no-createrole --pwprompt
 
 createdb suitec --owner=suitec
 createdb suitec_test --owner=suitec
+
+# Give SuiteC schema ownership on the test database so that it can reset the schema during tests
+psql suitec_test -c 'alter schema public owner to suitec;'
```
paulkerschen: Devs will probably need to run the line above on their local machines.

config/schema.sql
```diff
@@ -7,27 +7,11 @@
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
-SET idle_in_transaction_session_timeout = 0;
```
paulkerschen: This line is a Postgres 9.6 command that has no meaning for Postgres 9.5; it snuck in because of using pg_dump 9.6.

config/schema.sql
```diff
@@ -7,27 +7,11 @@
...
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
-
-
---
--- Name: EXTENSION plpgsql; Type: COMMENT; Schema: -; Owner:
---
-
-COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
-
-
```
paulkerschen: Attempting to create this extension led to permissions headaches, and doesn't appear to be necessary in any of our environments.

config/schema.sql
```diff
@@ -63,6 +47,7 @@ CREATE TYPE enum_activities_type AS ENUM (
     'comment',
     'get_comment',
     'get_asset_comment_reply',
+    'pin_asset',
```
paulkerschen: The proof is in the pudding: running the reference schema against actual code revealed that we'd missed this value.

johncrossman: Whoops!

johncrossman:
Looks good to me, @paulkerschen  We turn now to Travis.

PR #585, opened 2017-07-12
COL-1104, courses db table gets 'aws_s3_object_key'; eligible-for-S3 cutoff date in config

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1104

Proposal:
1. course `created_at` date must be greater than `aws.s3.suitecCutoverDate` to be eligible for AWS S3 (I will document this config if PR is merged)
1. `assets` table gets column for S3 Object Key
1. asset downloadUrl (for users) is concatenation of `aws.s3. baseDownloadUrl` (CloudFront) and S3 key.

If we're happy with this paradigm then future PRs will leverage (and test) the design.

### DIFF
--- a/node_modules/col-core/lib/db.js
+++ b/node_modules/col-core/lib/db.js
@@ -406,6 +406,8 @@ var setUpModel = function(sequelize) {
    * @property {String}       [pdf_url]               The PDF url of the asset
    * @property {Object}       [preview_metadata]      The preview metadata
    * @property {String}       [url]                   The url of the asset
+   * @property {String}       [aws_s3_object_key]     The Object Key of asset stored in AWS S3
+   * @property {String}       [download_url]          The url used by users to download asset (file)
    * @property {String}       [source]                The source of the asset
    * @property {String}       [body]                  The body of the asset
    * @property {Number}       likes                   The number of likes on the asset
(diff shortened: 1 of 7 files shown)


paulkerschen:
Looks like a good setup, @johncrossman; my one suggestion would be that `cutoverDate` is perhaps a sufficient name for the config key. Love it or hate it, there's no escaping that we're in a  SuiteC context here.

node_modules/col-core/lib/db.js
```diff
@@ -406,6 +406,8 @@ var setUpModel = function(sequelize) {
    * @property {String}       [pdf_url]               The PDF url of the asset
    * @property {Object}       [preview_metadata]      The preview metadata
    * @property {String}       [url]                   The url of the asset
+   * @property {String}       [aws_s3_object_key]     The Object Key of asset stored in AWS S3
+   * @property {String}       [download_url]          The Canvas url used by users to download asset (file)
```
paulkerschen: Thanks for noticing that we were missing `download_url`, @johncrossman; of course it doesn't have to be a Canvas URL and points to the ets-collabosphere bucket for newer assets.

johncrossman:
Thanks, @paulkerschen  Amendments in.

PR #590, opened 2017-07-13
COL-1112, AWS.S3 connector type (mock or real) is configurable; see aws-sdk-factory

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1112

My attempt at factory pattern for `aws-sdk` dependency. Tests and Travis run against mock-Canvas; we need the same for AWS S3.

Fyi: https://github.com/MathieuLoutre/mock-aws-s3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ config/*
 node_modules/*
 !node_modules/col-*
 
+# Mock Amazon S3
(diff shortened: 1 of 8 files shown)


paulkerschen:
Making a mockery of S3, are we, @johncrossman?

PR #591, opened 2017-07-13
COL-1111 Update source environment for data refresh script to suitec-prod

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1111

The script can be run on any environment, including dev machines, as long as source and destination configs are present.


johncrossman:
Refactor looks good, @paulkerschen  (Bash has no business parsing JSON.)

PR #592, opened 2017-07-13
COL-1113 Fix asset ownership check when adding to whiteboard

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1113

This facepalm of a bug is another no-test-coverage special.


johncrossman:
Le whoops! Thanks for the correction, @paulkerschen

PR #593, opened 2017-07-14
COL-747 Quiet AssertionErrors from mock Canvas

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-747

The AssertionErrors in question are real errors caused by the mock Canvas not getting requests in the order it expects. For some reason they were not actually causing the tests to fail, and were logged in NODE_ENV=travis but not NODE_ENV=test. I have no explanation for this behavior.


johncrossman:
Better a known unknown than an unknown unknown. +1, @paulkerschen

PR #595, opened 2017-07-18
COL-1116 Include impact_score and trending_score in db refresh script

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1116


johncrossman:
The expanded lexicon works for me.

PR #597, opened 2017-07-19
COL-1097, download assets via proxy to S3.getObject

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1097

We might circumvent the need for CloudFront by maintaining our own lightweight proxy. The AWS S3 module makes it easy to get a handle on the IO-stream and file metadata. If this approach is fruitful then we can eliminate `download_url` column after migration off of Canvas Files API.

### DIFF
--- a/node_modules/col-assets/lib/rest.js
+++ b/node_modules/col-assets/lib/rest.js
@@ -24,13 +24,14 @@
  */
 
 var _ = require('lodash');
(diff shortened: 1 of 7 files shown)


paulkerschen:
Thanks @johncrossman: having confirmed that big file streams don't seem to bother the rest of the stack, looks good.

PR #599, opened 2017-07-21
COL-1097, S3 object keys stored in asset.download_url, whiteboard.image_url during transition phase

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1097

After getting to know `whiteboard.image_url` I decided to ditch the `aws_s3_object_key` design and give dual purpose to `download_url` and `image_url`. These columns will store (1) canvas URL or (2) S3 object key. I believe this simplifies the code. (Once we are off the Canvas filesystem we might rename the columns.)

An immediate impact of this PR: `/assets/:assetId/download` for all downloads, including Canvas-based files. I've tested and it works (on localhost). Other tests done locally:
1. Preview generation of all sorts
1. Whiteboard publish and remix
1. For a specific course, direct asset files to Amazon S3 and then alter the `cutoverDate` config such that subsequent assets (of same course) are put to Canvas filesystem.  For this course, I was able to download files (S3-based and Canvas-based), create whiteboards, etc.

If this PR is merged I will create a separate Jira to remove `aws_s3_object_key` columns.


paulkerschen:
I checked the prod database to confirm that this is a safe distinction.

```
select count(*) from assets where download_url is not null;
> 97022
select count(*) from assets where download_url like 'http%';
> 97022
```

Thanks @johncrossman!

PR #600, opened 2017-07-21
COL-1121 Separate Mixpanel code and configs from overall analytics

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1121

Move over, Mixpanel, and make room for IMS Caliper, the standard of the future!

I've documented in release instructions a required change to local configs on suitec-qa and suitec-prod (but not on suitec-dev or developer machines, which are not hooked up to Mixpanel).

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -14,8 +14,10 @@
     }
   },
   "analytics": {
-    "enabled": true,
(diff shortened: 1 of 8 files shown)


johncrossman:
image

PR #601, opened 2017-07-21
COL-1136, upgrade aws-sdk; shuffle package.json

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1136

* reorder elements of package.json to put important bits on top (matches best practice, as far as I can tell) 
* The script change silences `npm install`

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,13 +6,24 @@
     "type": "git",
     "url": "git://github.com/ets-berkeley-edu/suitec.git"
   },
-  "bugs": {
-    "url": "https://jira.ets.berkeley.edu/jira/browse/COL"
-  },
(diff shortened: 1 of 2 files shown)


paulkerschen:
@johncrossman Have you tried an Elastic Beanstalk deploy with this revision? I'm curious whether shifting gulp and friends to devDependencies would cause npm not to install them under production mode and break the front-end build step. (See recent comments: https://stackoverflow.com/questions/20552373/how-does-aws-beanstalk-use-npm-when-deploying-a-nodejs-app)

johncrossman:
I've ditched the quite unessential `devDependencies`.  AWS upgrade remains.

PR #602, opened 2017-07-24
COL-1122 Add caliper-js dependency and wrapping analytics module

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1122

We work around the private status of the caliper.js library by (for now) installing from a fork under a developer account. Access is granted via https/API key rather than ssh so that both Travis and Elastic Beanstalk can use it. (Elastic Beanstalk downloads the key from S3; Travis has been given the key in an environment variable.)

With this change, developer machines, etc. wishing to run `npm install` will need to either copy the ~/.netrc file from S3 or use a Github account with "collaborator" status on the personal caliper-js fork.

### DIFF
--- a/node_modules/col-analytics/lib/api.js
+++ b/node_modules/col-analytics/lib/api.js
@@ -26,8 +26,10 @@
 var _ = require('lodash');
 var config = require('config');
 
+var CaliperAPI = require('./caliper');
 var MixpanelAPI = require('./mixpanel');
 
+var caliperEnabled = config.get('analytics.caliper.enabled');
 var mixpanelEnabled = config.get('analytics.mixpanel.enabled');
 
 /**
@@ -36,6 +38,9 @@ var mixpanelEnabled = config.get('analytics.mixpanel.enabled');
  * @return {void}
  */
 var init = function() {
+  if (caliperEnabled) {
+    CaliperAPI.init();
+  }
   if (mixpanelEnabled) {
     MixpanelAPI.init();
   }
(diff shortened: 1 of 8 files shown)


node_modules/col-analytics/lib/api.js
```diff
@@ -56,12 +61,15 @@ var identifyUser = module.exports.identifyUser = function(user) {
...
+ * @param  {Object}         [metadata]      Optional event metadata
  * @return {void}
  */
-var track = module.exports.track = function(user, event, options) {
+var track = module.exports.track = function(user, event, metadata) {
+  if (caliperEnabled) {
+    CaliperAPI.track(user, event, options);
+  }
   if (mixpanelEnabled) {
     MixpanelAPI.track(user, event, options);
```
johncrossman: @paulkerschen does `options` above need to be `metadata`?

node_modules/col-analytics/lib/caliper.js
```diff
@@ -0,0 +1,80 @@
...
+    log.error(err);
+  }
+};
+
+var getCourseIfMissing = function(user, callback) {
+  if (user.course) {
+    return callback(null, user.course);
+  }
+
+  CourseAPI.getCourse(user.course_id, callback);
```
johncrossman: @paulkerschen I'd expect `callback(null, CourseAPI.get...)` above.

node_modules/col-analytics/lib/api.js
paulkerschen: @johncrossman It does, thank you.

node_modules/col-analytics/lib/caliper.js
paulkerschen: The execution of the callback happens inside of CourseAPI.getCourse. This is just a shorthand to avoid having to write 

```
CourseAPI.getCourse(user.course_id, function(err, course) {
  return callback(err, course);
});
```

johncrossman:
Looks good, @paulkerschen I posted two questions. (This was not intended for new "caliper-experimental" branch, right?)

node_modules/col-analytics/lib/caliper.js
johncrossman: Of course – thank you.

PR #607, opened 2017-07-25
COL-1149, SuiteC stores S3 URIs in db (s3://bucket-name/my-object-key); not simply the object key

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1149

Having complete `s3://bucket-name/my-object-key` in the db is (1) more informative and (2) more obvious in purpose. We can pass these URIs directly to preview-service and get desired results.

Example of this URI scheme: 
http://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-plan-file-systems.html


johncrossman:
Sister PR of preview-service: https://github.com/ets-berkeley-edu/suitec-preview-service/pull/41

node_modules/col-core/lib/storage.js
```diff
@@ -162,21 +160,36 @@ var getObjectMetadata = module.exports.getObjectMetadata = function(key, callbac
...
+  return s3Uri.split('/').slice(3).join('/');
+};
+
+/**
+ * @param  {String}     s3Uri          URI of object in S3 (e.g., s3://my-bucket/my-object-key)
+ * @return {String}                    Strictly the S3 bucket name
+ */
+var extractS3Bucket = function(s3Uri) {
+  return s3Uri.split('/')[2];
 };
```
paulkerschen: @johncrossman This isn't the biggest deal in the world, but it strikes me that since `extractS3ObjectKey` and `extractS3Bucket` are always called together, perhaps they should just be one function? Then you'd only have to call `split('/')` once and could return both values.

johncrossman: Good advice, @paulkerschen  Amendment on its way

node_modules/col-core/lib/storage.js
```diff
@@ -162,21 +147,33 @@ var getObjectMetadata = module.exports.getObjectMetadata = function(key, callbac
...
+  return 's3://' + bucket + '/' + s3ObjectKey;
+};
+
+/**
+ * @param  {String}     s3Uri          URI of object in S3 (e.g., s3://my-bucket/my-object-key)
+ * @return {String}                    Strictly the S3 Object Key
  */
-var getUrl = module.exports.getUrl = function(uri) {
-  return isS3ObjectKey(uri) ? config.get('aws.s3.baseDownloadUrl') + '/' + uri : uri;
+var extractS3ObjectKey = function(s3Uri) {
```
paulkerschen: @johncrossman This function is now called `getS3Params`, yes?

johncrossman:
Whoops! Amendment is in. I verified locally w/ preview-service integration and arbitrary asset:
download-1

paulkerschen:
The land down under, where men blunder. Looking good!

PR #608, opened 2017-07-26
COL-573, fix failing file downloads (variable naming); remove obsolete baseDownloadUrl

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-573

If we resurrect the CloudFront design then `baseDownloadUrl` might return. (Storage code will get better test coverage soon.)


paulkerschen:
Right!

PR #612, opened 2017-07-31
COL-1128 Caliper instrumentation: asset (dis/un)liking

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1128

'Disliking' is of course not supported in the current UX, but we may as well shoot for consistency.

Like and dislike events generate a Caliper AnnotationEntity with id derived from the asset and user. 'Unliking' (i.e., removing a previous like or dislike) is then modeled as a Caliper 'remove' action taking that AnnotationEntity as its object.


johncrossman:
I neither dislike nor unlike this change. Thanks, @paulkerschen

PR #613, opened 2017-07-31
COL-573, perform auth check prior to asset download

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-573

You are not allowed to download asset if:
* not 'active'
* AND not admin user


paulkerschen:
Thanks for extra test coverage, @johncrossman; looks good.

PR #618, opened 2017-08-01
COL-1085, migration script > Canvas filesystem to Amazon S3

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1085

The migrate_assets script will download file to temp dir, upload to S3 and record resulting `s3://` address. The `mockFileUpload` op in our test utils did not produce a valid download_url therefore I extended the test util and our migrate script to recognize `file://` URIs. (Note: temp downloads are timestamped to avoid collisions.)


paulkerschen:
That migration business is more complicated than I remembered. Thanks, @johncrossman, all looks good to me.

PR #622, opened 2017-08-02
COL-1139 Avoid misleading dead-poller alert when no courses are active

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1139


johncrossman:
The reports of the poller death are greatly exaggerated. Thx, @paulkerschen

PR #623, opened 2017-08-03
COL-1106, script to move Canvas files to Amazon S3

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1106

Usage: `$0  --after [YYYY-MM-DD] --before [YYYY-MM-DD]` 

Canvas files are not removed after move is done – we don't want a bug in the script to cause a catastrophic loss of content. This script generates a CSV file which lists all "obsolete" files in Canvas. Once the move to Amazon S3 has been verified we can delete files in Canvas using these CSV(s).

### DIFF
--- a/scripts/move_canvas_files_to_amazon_s3.js
+++ b/scripts/move_canvas_files_to_amazon_s3.js
@@ -1,0 +1,327 @@
+#!/usr/bin/env node
+
+/**
+ * Copyright ©2017. The Regents of the University of California (Regents). All Rights Reserved.
+ *
+ * Permission to use, copy, modify, and distribute this software and its documentation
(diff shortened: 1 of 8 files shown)


scripts/move_canvas_files_to_amazon_s3.js
```diff
@@ -0,0 +1,327 @@
...
+  .alias('h', 'help')
+  .argv;
+
+/**
+ * Log message and prepend/append line breaks
+ *
+ * @param  {String}       message            What to log
+ * @return {void}
+ */
+var emphatic = function(message) {
```
paulkerschen: I detect emphasis!

paulkerschen:
Looks very nice, @johncrossman. Reporting rather than deleting is wise.

PR #625, opened 2017-08-04
COL-1159, if asset image is in Amazon S3 then use /download route to grab it

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1159

whiteboardDirective is not impressed with our half-baked `s3://` protocol.  We must construct URL as seen on asset detail page > download button.


johncrossman:
@paulkerschen should review because it's in the gnarly whiteboard front-end code.

paulkerschen:
Makes sense; thanks @johncrossman.

PR #630, opened 2017-08-14
COL-1165 S3 migration script: attempt to guard against duplicate callbacks

paulkerschen:
It's well-nigh impossible to reason through exactly what's going wrong with the async flow in https://jira.ets.berkeley.edu/jira/browse/COL-1165, but here's a shot at clearing it up.

### DIFF
--- a/scripts/move_canvas_files_to_amazon_s3.js
+++ b/scripts/move_canvas_files_to_amazon_s3.js
@@ -190,9 +190,11 @@ var moveFilesToAmazonS3 = function(course, callback) {
     log.info({'course': course.id}, assets.length + ' assets in course \'' + course.name + '\'');
 
     async.each(assets, function(asset, done) {
+      var doneOnce = _.once(done);
+
       if (Storage.isS3Uri(asset.download_url) || !_.startsWith(asset.download_url, 'http')) {
         // No move is necessary for this asset
-        return done();
+        return doneOnce();
 
       } else {
         log.info({'asset': asset.id}, 'Move ' + asset.title);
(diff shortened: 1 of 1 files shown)


paulkerschen:
Where'd you find the docs on the stream 'error' event, @raydavis? I went looking and didn't turn them up.

raydavis:
It's frustratingly quiet about how one _should_ handle 'error' events, but: https://nodejs.org/api/stream.html#stream_event_error

A couple of other links document struggles with getting stream error handling to do something worthwhile:
* https://www.bennadel.com/blog/2679-how-error-events-affect-piped-streams-in-node-js.htm
* http://grokbase.com/t/gg/nodejs/12bwd4zm4x/should-stream-pipe-forward-errors
* https://github.com/ludios/why-are-my-node-streams-hanging

I'm not sure what the right thing to do is in this case, but the current code doesn't look like it.

paulkerschen:
Thanks for the links, @raydavis. The perplexities of the authors seem to match my unhappy experience with Node stream APIs, which is that you never quite know whether an explicit 'finish' is going to get called after the 'error'. In this context, it may actually be the safest thing to have them both execute the callback; with the proviso that in the context of an async loop, the callback must execute once and once only, hence wrapping it in a `_.once` safety belt within that loop. Does that seem plausible enough to you?

raydavis:
Although that might stymie the double-callback, wouldn't it also be a good idea to log the error?

paulkerschen:
The error should be passed back from line 143 to line 220, and logged on line 222.

raydavis:
Ah, I see -- the existing log messages are whipped cream on top of a scoop of vanilla logging.

PR #633, opened 2017-08-16
COL-1166 Tell socket.io not to attempt failing websocket handshake

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1166

It's unknown why our websocket handshake is currently failing on AWS, but until we get around to investigating, we may as well tell our code not to attempt and error.


raydavis:
Not very bleeding-edge of AWS, is it?

PR #634, opened 2017-08-22
COL-1178 Do not display sync option for assigments without syncable submission types

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1178

PR #636, opened 2017-08-23
COL-1178 Short-circuit submission processing for non-syncable assignments

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1178

More async control flow (sigh) as a follow up to #634, which broke suitec-dev.


johncrossman:
+1  Thanks, @paulkerschen

PR #645, opened 2017-08-29
COL-1182, assert.ifError(err) will give error details if err

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1182

For example, in a `recalculateImpactScores` test I added:
```
assert.ifError(new Error('something bad happened'));
```
The error message surfaced after Mocha had run _all_ tests.
image


paulkerschen:
A monotonous yet strangely reassuring diff. Thanks @johncrossman.

PR #651, opened 2017-09-07
COL-1183 Failure to find activity during delete request is no cause for alarm

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1183

Keep calm. ActivitiesAPI.deleteActivity is used to remove any previous activities that may match certain criteria - for example, as part of creating a like activity, any previous likes are removed. Failure to find a match is quite in the normal course of business and doesn't need to be logged.


johncrossman:
Keep calm and impact on. Thx, @paulkerschen  Waiting on T

PR #653, opened 2017-09-20
COL-1196, video_url, on asset detail view, is preview_metadata.converted_video or /assets/:id/download

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1196

When it comes to `download_url` and `video_url`, SuiteC server-side is a proxy to Amazon S3.


raydavis:
The eavesdropping made sense to me...

PR #654, opened 2017-09-21
COL-1194, hidden assets can be viewed by only collaborators and admins

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1194

Putting this logic at `/asset/:asset_id` and `/download/...` was a bit tougher than expected. I use a db query instead of WhiteboardsAPI call, which I believe would be more expensive. (See `badgeEligibleAssetsQuery` for similar decision.)

### DIFF
--- a/node_modules/col-assets/lib/api.js
+++ b/node_modules/col-assets/lib/api.js
@@ -1145,15 +1145,65 @@ var deleteAssets = module.exports.deleteAssets = function(ctx, assetIds, callbac
 };
 
 /**
+ * @param   {Category[]}    categories       The categories to review
+ * @return  {boolean}                        True if empty or at least one category is visible
+ */
+var isEmptyOrHasVisible = function(categories) {
+  return _.isEmpty(categories) || _.find(categories, 'visible');
+};
+
+/**
  * Returns true if the asset is hidden, deleted, or associated with only hidden categories.
  *
  * @param   {Asset}        asset             The asset to check for deleted or hidden status
  * @return  {boolean}                        True if deleted or hidden
  */
 var isDeletedOrHidden = module.exports.isDeletedOrHidden = function(asset) {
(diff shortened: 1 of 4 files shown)


node_modules/col-assets/lib/api.js
```diff
@@ -1154,6 +1154,48 @@ var isDeletedOrHidden = module.exports.isDeletedOrHidden = function(asset) {
   return !asset.visible || !!asset.deleted_at || !(_.isEmpty(asset.categories) || _.find(asset.categories, 'visible'));
 };
 
+/**
+ * Returns true user is authorized to view the specified asset.
```
paulkerschen: @johncrossman missing `if`?

node_modules/col-assets/lib/api.js
```diff
@@ -1154,6 +1154,48 @@ var isDeletedOrHidden = module.exports.isDeletedOrHidden = function(asset) {
...
+ */
+var canUserView = module.exports.canUserView = function(user, asset) {
+  var canUserView = false;
+
+  if (asset.deleted_at) {
+    canUserView = false;
+  } else if (user.is_admin || _.find(asset.users, {'id': user.id})) {
+    canUserView = true;
+  } else {
+    var hidden = !asset.visible || (!_.isEmpty(asset.categories) && !_.find(asset.categories, 'visible'));
```
paulkerschen: @johncrossman Since this compound Boolean is close to the one in `isDeletedOrHidden`, above, I wonder if we should move it into its own `isHidden` method and refer to that method in both places?

paulkerschen:
Looks good, @johncrossman - I agree that hitting the db directly for whiteboards makes sense. Posted a couple of questions.

node_modules/col-assets/lib/api.js
johncrossman: @paulkerschen To preserve the (near) readability, I extracted only the `categories` logic. PR amended.

PR #655, opened 2017-09-22
COL-1194, collaborator can see whiteboard assets put by other collaborator

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1194

I've been bitten by idiosyncratic async.  I recommend `?w=1`


paulkerschen:
Callbacks make it twice as nice. Thanks @johncrossman.

PR #656, opened 2017-09-28
COL-1108, post S3-migration: remove code associated with file storage in Canvas

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1108

* the `move_canvas_files_to_amazon_s3.js` script is still usable
* Canvas assignment integration should not be impacted
* all obsolete code removed, I hope


paulkerschen:
Thanks @johncrossman - what a satisfying pink diff.

PR #659, opened 2017-09-28
COL-1201, upgrade aws-sdk to 2.124.0

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1201

https://github.com/aws/aws-sdk-js/blob/master/CHANGELOG.md

Snuck in minor `eslint` and `moment` upgrades, too.


paulkerschen:
Indeed, the future is here.

PR #661, opened 2017-10-16
NOJIRA, refresh-data.sh > exit if dest_host URL contains '-prod'

johncrossman:
In the fast-paced world of copy-and-paste, I don't want prod configs to accidentally land in `${dest_suitec}.json`.


paulkerschen:
Thanks @johncrossman; this will work very well until we change the service name to `cattle-prod`.

PR #662, opened 2017-10-20
COL-1212 Enable session stickiness in .ebextensions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1212

http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environments-cfg-applicationloadbalancer.html


johncrossman:
Sticky icky!  Thx, @paulkerschen

PR #663, opened 2017-10-23
COL-1217, Canvas clean up script now accepts canvas_course_id arg

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1217

Usage info says it all.

### DIFF
--- a/scripts/20170928-col1108/clean_up_canvas_after_s3_migration.js
+++ b/scripts/20170928-col1108/clean_up_canvas_after_s3_migration.js
@@ -47,13 +47,15 @@ var timezone = config.get('timezone');
 var totalCourseCount = 0;
 
 var argv = yargs
-  .usage('Usage: node $0 --after [YYYY-MM-DD] --before [YYYY-MM-DD] --csv_directory /writable/dir/for/csv-files/')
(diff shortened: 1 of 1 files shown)


scripts/20170928-col1108/clean_up_canvas_after_s3_migration.js
```diff
@@ -47,13 +47,15 @@ var timezone = config.get('timezone');
 var totalCourseCount = 0;
 
 var argv = yargs
-  .usage('Usage: node $0 --after [YYYY-MM-DD] --before [YYYY-MM-DD] --csv_directory /writable/dir/for/csv-files/')
+  .usage('Usage: node $0 [--after YYYY-MM-DD] [--before YYYY-MM-DD] [--canvas_course_id 123] --csv_directory /writable/dir/for/csv-files/')
   .demand([ 'csv_directory' ])
   .describe('after', 'Courses created after this date are eligible')
   .describe('before', 'Courses created before this date are eligible')
+  .describe('canvas_course_id', 'The Canvas id of course to process')
```
paulkerschen: @johncrossman Would it make sense to use `id` (which is opaque but unique) rather than `canvas_course_id` (which is more apparent in the UI, but could potentially match multiple courses in different Canvas instances)?

johncrossman: Good point, @paulkerschen  Mod coming soon

PR #664, opened 2017-10-25
COL-1218 Make forever truly forever

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1218

Rather than put a cap on the number of restarts, always restart unless the app stays up for less than 10 seconds.


johncrossman:
image

PR #667, opened 2017-11-06
COL-1223 Student-to-student activities API feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1223


johncrossman:
Top-to-bottom, student-to-student... Looks good

PR #668, opened 2017-11-06
COL-1233, clean_up_canvas script looks for '_collabosphere' folder if '_suitec' was not found

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1233

If neither folder is found then report the 404 to 'notFound' file.


paulkerschen:
Clean-up-Canvas sheriff is in town.

PR #669, opened 2017-11-06
COL-1236, clean_up_canvas script > increase timeout for delete requests

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1236

Read about `timeout` option and read timeouts at https://github.com/request/request 

We leverage `canvasPoller.timeout` config


paulkerschen:
Exactly.

PR #673, opened 2017-11-13
COL-1241 Exclude inactive users from interactions feed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1241

These users don't show up in the user feed, meaning that d3-force can't match them up and goes on to lose its mind.


johncrossman:
Keep d3 sane! Thanks, @paulkerschen

PR #675, opened 2017-11-14
COL-1226 Activity network interaction details tooltip

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1226

Note the expanded canvas_full_name for the selected user (useful in a course site with two Pauls, for example).

screen shot 2017-11-14 at 12 00 08 pm

screen shot 2017-11-14 at 12 00 28 pm


johncrossman:
Neither tool nor tip out of place. Thanks, @paulkerschen

PR #678, opened 2017-11-16
COL-1229 Cropping and panning for activity network

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1229

An approach in progress to deal with the size of the W4C network. Adding pan behavior entails losing the (fun but useless) drag behavior where you could grab and bounce individual people's heads at will.


johncrossman:
From this viewport, all looks good. Waiting on T.

PR #681, opened 2017-11-17
COL-1229 Activity Network: text changes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1229

Also HTML ids on user node elements for QA dept.


johncrossman:
Words in papers, words in books
Words on TV, words for crooks...

PR #683, opened 2017-11-17
COL-1249 One is the loneliest number

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1249

Includes a tapering function for line thickness, now that we have properly large numbers.


johncrossman:
And it takes two to make a thing go right. +1, @paulkerschen

PR #685, opened 2017-11-20
COL-1247 Activity Network: refocusable user nodes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1247


johncrossman:
Looks right. Waiting on T

PR #686, opened 2017-11-20
COL-1247 Activity network: follow-up boundary handling for focal user

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1247

Follow-up to #685. `scope.user` is constrained to appear near the center of the diagram. `scope.focalUser` does not need this constraint; it just needs the larger bubble size accounted for.


johncrossman:
Respect for the boundaries.

PR #698, opened 2018-01-18
COL-1266 Update self-signed instance certificate

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1266

This updated certificate is good for the next year, through 18 Jan 2019. Calendaring these updates for the future will be a follow-up.


johncrossman:
Um, okay... Sure.  Thanks, @paulkerschen

PR #702, opened 2018-02-08
COL-1283, create db table to track 'events' a la MixPanel

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1283
* I chose not to include foreign key constraints. This is an audit log and should not 'cascade on delete' or similar.
* The not-null columns: event_name, event_metadata, canvas_domain

### DIFF
--- a/scripts/20180208-col1283/create_events_table.sql
+++ b/scripts/20180208-col1283/create_events_table.sql
@@ -1,0 +1,35 @@
+-- What was sent to MixPanel will now be written to 'events' table
+
+CREATE TABLE events (
+  id SERIAL PRIMARY KEY,
+  uuid uuid NOT NULL,
+  event_name VARCHAR(255) NOT NULL,
+  event_metadata jsonb NOT NULL,
+  canvas_domain VARCHAR(255) NOT NULL,
+  user_id INTEGER,
+  canvas_user_id INTEGER,
+  canvas_full_name VARCHAR(255),
+  canvas_course_role VARCHAR(255),
+  course_id INTEGER,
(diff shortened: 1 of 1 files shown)


paulkerschen:
Looks good from my end, @johncrossman, but @sandeepmjay might want to review- I think he was hoping for a generated UUID string.

johncrossman:
Thanks @paulkerschen  I've amended to include `uuid` (with type uuid). I considered it as primary-key but, considering [the pitfalls](https://tomharrisonjr.com/uuid-or-guid-as-primary-keys-be-careful-7b2aa3dcb439), I decided against.  cc: @sandeepmjay

johncrossman:
@sandeepmjay what should change, if anything, in this db table def?

sandeepmjay:
@johncrossman @paulkerschen 
Mixpanel event feed uses only 2 keys {event: string, properties:object}. Storing mixpanel event properties as event.metadata is a good idea. It would make it easier to re-stream events back to mixpanel if researchers got with it as a tool of choice.   User metadata seems unnecessary if we have a user_id. 

I recommend parsing any object_ids(assets, comments, whiteboards) that we come across in the properties and adding them as columns for easier filtering on the event log table.

johncrossman:
@sandeepmjay @paulkerschen Amendment in, Travis green. I removed `user_metadata` and added nullable columns: `asset_id`, `comment_id`, `whiteboard_id`, `whiteboard_element_uid`. (The latter `_uid` suffix matches naming in whiteboard_elements table.

sandeepmjay:
Now, that looks like a mixpanel killer.

PR #704, opened 2018-02-09
COL-1281 Apply stricter permissions to config/local-production.json

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1281


johncrossman:
Bigger, better, stricter!

PR #706, opened 2018-02-13
COL-1286, event-types are categorized based on primary object association (eg, asset)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1286

The `events` table gets object ids, when available. Event types are categorized with hope of a sensible switch statement. Tests are in next PR.

### DIFF
--- a/node_modules/col-analytics/lib/util.js
+++ b/node_modules/col-analytics/lib/util.js
@@ -23,7 +23,62 @@
  * ENHANCEMENTS, OR MODIFICATIONS.
  */
 
+var _ = require('lodash');
+
 var CourseAPI = require('col-course');
+var EventTypes = require('./constants');
+
+/**
+ * Certain object metadata is static and available elsewhere in the SuiteC schema. For example,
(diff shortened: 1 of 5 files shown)


node_modules/col-analytics/lib/util.js
```diff
@@ -23,7 +23,64 @@
...
+  'asset_dislikes',
+  'asset_image_url',
+  'asset_liked',
+  'asset_mime',
+  'asset_source',
+  'asset_thumbnail_url',
+  'asset_type',
+  'asset_url',
+  'asset_users',
+  'comment_body',
```
paulkerschen: @johncrossman Is there any value to including `comment_body` and `asset_description` if they might be edited over time, or do you think they're just too lengthy and noisy?

johncrossman: @paulkerschen  I excluded them for reasons you mention but they do qualify as mutable. I'll remove from blacklist.

paulkerschen:
What a lot of data!

PR #707, opened 2018-02-13
COL-1291, do not record 'Get engagement index' event when user hits profile page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1291

On server side, at `/users/leaderboard`, we have:
```
  // Track the request unless explicitly told otherwise.
  var track = true;
  if (CollabosphereUtil.getBooleanParam(req.query.track) === false) {
    track = false;
  }
   ...
```


paulkerschen:
Ah, a rare straightforward bug. Thanks @johncrossman

PR #709, opened 2018-02-14
COL-1285, verify event when create whiteboard and create element

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1285

Last `event` tracking test, for now.


paulkerschen:
I only wish that testing elements was elementary. Thanks @johncrossman.

PR #712, opened 2018-02-16
COL-1277 Provide optional asset id to preview reprocessing script

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1277


johncrossman:
It's good to have options.

PR #714, opened 2018-02-16
COL-1301, rest endpoint to track events, accepting optional id params

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1301

In preparation for client-side cutover, we make `/api/track` available. The long list of optional object-id args seems cleaner than a more obscure `object` and `contextObject`.

### DIFF
--- a/node_modules/col-analytics/tests/util.js
+++ b/node_modules/col-analytics/tests/util.js
@@ -240,15 +240,16 @@ var getMostRecentEvent = module.exports.getMostRecentEvent = function(course, us
 /**
  * Assert that Event has expected properties
  *
- * @param  {Object}         event                      The actual event received in test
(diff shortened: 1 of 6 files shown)


node_modules/col-analytics/tests/util.js
```diff
@@ -289,5 +290,39 @@ var assertEvent = module.exports.assertEvent = function(event, eventName, user,
...
+  eventName,
+  metadata,
+  activityId,
+  assetId,
+  commentId,
+  whiteboardId,
+  whiteboardElementUid,
+  callback
+) {
+  // Too much of a good thing
```
paulkerschen: Never too much!

paulkerschen:
Thanks, @johncrossman, I agree that the long list of optional parameters is a lesser evil.

PR #715, opened 2018-02-18
COL-1297, remove client-side mixpanel.track(); scrub metadata

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1297
https://jira.ets.berkeley.edu/jira/browse/COL-1299

We do our best to extract object ids. And, remove extraneous and duplicative info from event_metadata object.


johncrossman:
@paulkerschen is on vacay. I might need a review from @raydavis or @sandeepmjay

raydavis:
Our best looks OK to me.

PR #722, opened 2018-02-26
COL-1314, analyticsService.track invokes async $q wrapping trackEvent

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1314

As promised, I briefly investigated "drops" in event tracking (eg, EI searches). This PR adds `$q` usage ([Angular service](https://docs.angularjs.org/api/ng/service/$q) runs functions async) which seems reliable in local testing. This might well reduce "drops".

### DIFF
--- a/public/app/shared/analyticsService.js
+++ b/public/app/shared/analyticsService.js
@@ -27,7 +27,7 @@
 
   'use strict';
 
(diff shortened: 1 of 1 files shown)


johncrossman:
cc: @paulkerschen

paulkerschen:
Of course, $q always builds the most amazing gadgets.

PR #724, opened 2018-03-06
COL-1321 Use Canvas Conversations UI for user messaging

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1321

Given masquerade permission issues, don't try to create Canvas conversation messages via the API. Redirecting to the Canvas Conversations UI lets us dump all that scaffolding.


johncrossman:
Such a beautiful sea of deletions. Waiting on T

PR #725, opened 2018-03-27
COL-1324 Give that Activity Network a border

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1324


johncrossman:
Make love, not borders!

PR #726, opened 2018-03-27
COL-1328 Activity Network: Enable user details popup on big center node

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1328

`?w=1` will reveal there's less here than meets the eye.


johncrossman:
I am he, as you are me, and we are all together.

PR #727, opened 2018-03-28
COL-1325 Activity network: recent/all toggle

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1325

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -11,6 +11,9 @@
       "version": "2006-03-01"
     }
   },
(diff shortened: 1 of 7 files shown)


johncrossman:
@paulkerschen - Looks good but Travis would like a word...

paulkerschen:
@johncrossman I repeated myself to Travis, more slowly and louder, and it got through.

PR #731, opened 2018-04-02
[QA] COL-1329, call the filter "Last 90 Days" instead of "Recent"

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1329


johncrossman:
PR amended. The day count displayed reflects value of `activityNetwork.recentUserCutoff` config.

PR #736, opened 2018-05-14
COL-1333 Please don't destructively modify the UserConstants

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1333


johncrossman:
There is nothing constant except sneaky bugs. Thx @paulkerschen

PR #760, opened 2019-03-20
COL-1380 Re-sign image preview URLs as needed

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1380

### DIFF
--- a/node_modules/col-core/lib/storage.js
+++ b/node_modules/col-core/lib/storage.js
@@ -32,6 +32,7 @@ var mime = require('mime');
 var moment = require('moment-timezone');
 var path = require('path');
 var timezone = config.get('timezone');
+var url = require('url');
 var util = require('util');
 
 var s3 = AWS.S3();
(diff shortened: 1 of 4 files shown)


node_modules/col-core/lib/storage.js
```diff
@@ -176,9 +223,6 @@ var putObjectToS3 = function(key, filePath, callback) {
       return callback(err);
     }
 
-    // Generate a signed download URL
-    var expires = 2 * 365 * 24 * 60 * 60;
-
```
paulkerschen: Strangely, we have here an unused cut-and-paste from the preview-service code that started the whole problem.

paulkerschen:
Travis integration on this repo seems to have withered with disuse, but `gulp test` passes for me locally.

johncrossman:
"Withered with disuse" threatens to be a tagline. Looks good, @paulkerschen

PR #767, opened 2019-04-18
[QA] COL-1391, ignores active submission if attachment is in pending_upload state

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1391


paulkerschen:
Trickier and trickier.

PR #772, opened 2019-05-29
COL-1389, upgrade Sequelize to v5.8.7

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1389

I'll create a less urgent follow up Jira to do away with 'operatorsAliases'.  See http://docs.sequelizejs.com/manual/querying.html#operators-security


paulkerschen:
Roll forward and forward again.

PR #774, opened 2019-06-12
COL-1395, bower/gulp belongs in buildspec – not ebextensions

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1395

The build step (bower, gulp) does not belong in the deploy phase.


johncrossman:
This change is motivated by mysterious failures during ec2 deploy phase. This change would have been a good idea, regardless.

PR #775, opened 2019-06-12
NOJIRA, SuiteC master branch bumped to v2.10.0

johncrossman:
The only thing better than a `package.json`  is 14 of them.


paulkerschen:
SuiteC's packages always ship bulk rate.

PR #785, opened 2020-01-31
COL-1410 Update cookie properties for Chrome 80

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1410
https://jira.ets.berkeley.edu/jira/browse/CLC-7207


johncrossman:
Thanks @paulkerschen   I did a few greps and it seems you caught all cases.

PR #786, opened 2020-01-31
COL-1398, no more eslint and bower security vulnerabilities, for now

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/COL-1398


paulkerschen:
The SHIELD!

PR #801, opened 2021-02-03
COL-1436 Yum is not yummy without -y

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1436


lyttam:
"Um" just won't do.

johncrossman:
Blech!

